PR #2, opened 2021-04-16
Create consistent hashing library based on a published paper

tcsnfkx:
Add implementation of consistent hashing based on https://arxiv.org/pdf/1406.2294.pdf.

The impelmentation is copied and formatted from Figure 1 on page 2 in https://arxiv.org/pdf/1406.2294.pdf.

### DIFF
--- a/src/main/cc/wfa/virtual_people/core/model/utils/consistent_hash.cc
+++ b/src/main/cc/wfa/virtual_people/core/model/utils/consistent_hash.cc
@@ -22,7 +22,7 @@ namespace wfa_virtual_people {
 // published paper:
 //   https://arxiv.org/pdf/1406.2294.pdf
 int32_t JumpConsistentHash(uint64_t key, int32_t num_buckets) {
-  int64_t b = 1, j = 0;
+  int64_t b = -1, j = 0;
   while (j < num_buckets) {
     b = j;
     key = key * 2862933555777941757ULL + 1;
--- a/src/main/cc/wfa/virtual_people/core/model/utils/consistent_hash.cc
+++ b/src/main/cc/wfa/virtual_people/core/model/utils/consistent_hash.cc
@@ -23,6 +23,7 @@ namespace wfa_virtual_people {
 //   https://arxiv.org/pdf/1406.2294.pdf
 int32_t JumpConsistentHash(uint64_t key, int32_t num_buckets) {
   int32_t b = -1;
+  // Use int64_t for j to handle int32_t overflow.
   for (int64_t j = 0; j < num_buckets; ) {
     b = j;
     key = key * 2862933555777941757ULL + 1;
--- a/src/main/cc/wfa/virtual_people/core/model/utils/consistent_hash.cc
+++ b/src/main/cc/wfa/virtual_people/core/model/utils/consistent_hash.cc
@@ -29,7 +29,7 @@ int32_t JumpConsistentHash(uint64_t key, int32_t num_buckets) {
     j = (b + 1) * (static_cast<double>(1LL << 31) /
                    static_cast<double>((key >> 33) + 1));
   }
-  return b;
+  return static_cast<int32_t>(b);
 }
 
 }  // namespace wfa_virtual_people
--- a/src/main/cc/wfa/virtual_people/core/model/utils/consistent_hash.cc
+++ b/src/main/cc/wfa/virtual_people/core/model/utils/consistent_hash.cc
@@ -1,0 +1,25 @@
+#include "wfa/virtual_people/core/model/utils/consistent_hash.h"
+
+#include <cstdint>
+
+namespace wfa_virtual_people {
+namespace core {
+namespace utils {
+
+// This implementation is a copy and formatting of Figure 1 on page 2 in the
+// published paper:
+//   https://arxiv.org/pdf/1406.2294.pdf
+int32_t JumpConsistentHash(uint64_t key, int32_t num_buckets) {
+  int64_t b = 1, j = 0;
+  while (j < num_buckets) {
+    b = j;
+    key = key * 2862933555777941757ULL + 1;
+    j = (b + 1) * (static_cast<double>(1LL << 31) /
+                   static_cast<double>((key >> 33) + 1));
+  }
+  return b;
+}
+
+}  // namespace utils
+}  // namespace core
+}  // namespace wfa_virtual_people
--- a/src/main/cc/wfa/virtual_people/core/model/utils/consistent_hash.cc
+++ b/src/main/cc/wfa/virtual_people/core/model/utils/consistent_hash.cc
@@ -1,10 +1,22 @@
+// Copyright 2021 The Cross-Media Measurement Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
 #include "wfa/virtual_people/core/model/utils/consistent_hash.h"
 
 #include <cstdint>
 
 namespace wfa_virtual_people {
-namespace core {
-namespace utils {
 
 // This implementation is a copy and formatting of Figure 1 on page 2 in the
 // published paper:
@@ -20,6 +32,4 @@ int32_t JumpConsistentHash(uint64_t key, int32_t num_buckets) {
   return b;
 }
 
-}  // namespace utils
-}  // namespace core
 }  // namespace wfa_virtual_people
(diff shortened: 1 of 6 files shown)


src/main/cc/wfa/virtual_people/core/model/utils/consistent_hash.cc
```diff
@@ -0,0 +1,25 @@
+#include "wfa/virtual_people/core/model/utils/consistent_hash.h"
```
wliue, line 1: Add license header.

src/main/cc/wfa/virtual_people/core/model/utils/consistent_hash_test.cc
```diff
@@ -0,0 +1,48 @@
+#include "wfa/virtual_people/core/model/utils/consistent_hash.h"
```
wliue, line 1: Add license header.

src/main/cc/wfa/virtual_people/core/model/utils/consistent_hash.cc
tcsnfkx: Done.

src/main/cc/wfa/virtual_people/core/model/utils/consistent_hash_test.cc
tcsnfkx: Done.

src/main/cc/wfa/virtual_people/core/model/utils/consistent_hash.cc
```diff
@@ -0,0 +1,35 @@
...
+
+#include <cstdint>
+
+namespace wfa_virtual_people {
+
+// This implementation is a copy and formatting of Figure 1 on page 2 in the
+// published paper:
+//   https://arxiv.org/pdf/1406.2294.pdf
+int32_t JumpConsistentHash(uint64_t key, int32_t num_buckets) {
+  int64_t b = 1, j = 0;
```
lenjoy, line 25: `b = -1` in the paper, but apparently this doesn't matter

tcsnfkx: Fixed.

src/main/cc/wfa/virtual_people/core/model/utils/consistent_hash.cc
```diff
@@ -0,0 +1,35 @@
...
+
+// This implementation is a copy and formatting of Figure 1 on page 2 in the
+// published paper:
+//   https://arxiv.org/pdf/1406.2294.pdf
+int32_t JumpConsistentHash(uint64_t key, int32_t num_buckets) {
+  int64_t b = -1, j = 0;
+  while (j < num_buckets) {
+    b = j;
+    key = key * 2862933555777941757ULL + 1;
+    j = (b + 1) * (static_cast<double>(1LL << 31) /
```
lenjoy, line 29: i know this is from the code in the paper. from c++ perspective, we only need  `uint64` and `double` for the `1LL<<31 / key>>33 ...` only, how about making some slight change as the following:

```
int32_t b = -1;
for (int32_t j = 0; j < num_bucktes; ) {
  b = j;
  ...
  j = static_cast<int32_t>( ... );
}
```

tcsnfkx: Though this is very unlikely to happen, if num_bucktes is std::numeric_limits<int>::max, j < num_buckets will never be fulfilled if j is int32_t.

lenjoy: i meant, just to keep the type consistent in the code.   sure, we don't need to worry about the overflow

tcsnfkx: I think int32_t does not work for this algorithm. Changed to cast the return to int32_t explicitly.

lenjoy: don't change the return type here. see my code snippet above, `j = static_cast<int32_t>(...)` is the key part

tcsnfkx: I mean we cannot cast j to int32_t. If we cast j to int32_t, and the num_buckets is very big, the condition "j >= num_bucktes" is hard to fulfill, and the while/for loop will last too long. Furthermore, if num_buckets = INT_MAX, the while/for loop will never end.

lenjoy: what's the difference between int32 and int64 for `j` here?   anyway, j needs to reach `num_buckets` in jumps

tcsnfkx: "j needs to reach num_buckets in jumps", this is what I mean. j should be able to reach larger than maximum possible value of num_buckets.

lenjoy: the code can be like this:
```
// Generates the bucket index in range of [0, numBuckets).                                                                              
int32_t JumpConsistentHash(uint64_t key, int32_t num_buckets) {
  int32_t b = -1;
  for (int32_t j = 0; j < num_buckets; ) {
    b = j;
    key = key * 2862933555777941757ULL + 1;
    j = (j+1) * (static_cast<double>(1LL << 31) /
                 static_cast<double>((key >> 33) + 1));
  }
  return b;
}
```

tcsnfkx: How about
```
int32_t JumpConsistentHash(uint64_t key, int32_t num_buckets) {
  int32_t b = -1;
  for (int64_t j = 0; j < num_buckets; ) {
    b = j;
    key = key * 2862933555777941757ULL + 1;
    j = (j+1) * (static_cast<double>(1LL << 31) /
                 static_cast<double>((key >> 33) + 1));
  }
  return b;
}
```
Though j will reach num_buckets finally, the function is very slow when num_buckets is big. We will call this function for every event, so the performance will have great impact.

tcsnfkx: Just changed b from int64_t to int32_t, and still keep j as int64_t due to performance consideration.

lenjoy: `int64_t` doesn't look that rigorous.  how about just adding "j>=0" to prevent overflow?

```
// Generates the bucket index in range of [0, numBuckets).                                                                              
int32_t JumpConsistentHash(uint64_t key, int32_t num_buckets) {
  int32_t b = -1;
  for (int32_t j = 0; j < num_buckets && j >= 0; ) {
    // j>=0 to prevent overflow
    b = j;
    key = key * 2862933555777941757ULL + 1;
    j = (j+1) * (static_cast<double>(1LL << 31) /
                 static_cast<double>((key >> 33) + 1));
  }
  return b;
}
```

src/main/cc/wfa/virtual_people/core/model/utils/consistent_hash_test.cc
```diff
@@ -0,0 +1,58 @@
...
+constexpr int32_t kMaxBuckets = 1 << 16;
+
+// When using the same key, for any number of buckets n > 1, one of the
+// following must be satisfied
+// * JumpConsistentHash(key, n) == JumpConsistentHash(key, n - 1)
+// * JumpConsistentHash(key, n) == n - 1
+void CheckCorrectnessForOneKey(const uint64_t key, const int32_t max_buckets) {
+  int32_t last_output = 0;
+  for (int32_t num_buckets = 1; num_buckets <= max_buckets; ++num_buckets) {
+    int32_t output = JumpConsistentHash(key, num_buckets);
```
lenjoy, line 37: how about adding a simple test case with `num_buckets` as max int32?

lenjoy (changes requested):
LGTM in general, just some minor stuff

src/main/cc/wfa/virtual_people/core/model/utils/consistent_hash.cc
tcsnfkx: In the code
```
    j = (j+1) * (static_cast<double>(1LL << 31) /
                 static_cast<double>((key >> 33) + 1));
```
The value of right side is a double floating number. In c++, the conversion from floating number to integer with overflow is unidentified. So it is dangerous to assume that the overflow will cause j to be negative.

I think the behavior of setting j as int64_t was verified, as it is used in the code from the published paper. Changing the type to int32_t will cause the return to be different.

src/main/cc/wfa/virtual_people/core/model/utils/consistent_hash_test.cc
tcsnfkx: Test case added.

src/main/cc/wfa/virtual_people/core/model/utils/consistent_hash.cc
lenjoy: Makes sense! LGTM! Add a comment “int64 j to handle int32 overflow”

lenjoy (approved):

wliue (approved):

src/main/cc/wfa/virtual_people/core/model/utils/consistent_hash.cc
tcsnfkx: Comment added. Thanks!

PR #6, opened 2021-06-01
Implement BranchNode.

### LINKED ISSUE #31: Implement virtual people model and labeler in C++

tcsnfkx:
Create C++ implemention for BranchNode. Seleting child node by chance or condition is supported currently. Attributes updater and multipilicity will be implemented in future.

Add a method in ModelNode to resolve the child node index references.

---
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/6)

### DIFF
--- a/src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc
+++ b/src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc
@@ -149,13 +149,12 @@ BranchNodeImpl::BranchNodeImpl(
 absl::Status ResolveChildReference(
     ChildNodeRef& child_node,
     absl::flat_hash_map<uint32_t, std::unique_ptr<ModelNode>>& node_refs) {
-  if (child_node.index() == 0) {
+  if (uint32_t* node_index = std::get_if<0>(&child_node)) {
     // The child node is referenced by node index. Need to resolve to the
     // ModelNode object.
-    uint32_t node_index = std::get<0>(child_node);
     // The owner of the corresponding ModelNode pointer will be its parent node.
     // Gets the key value pair, and deletes them from the map.
-    auto nh = node_refs.extract(node_index);
+    auto nh = node_refs.extract(*node_index);
     if (nh.empty()) {
       return absl::InvalidArgumentError(
           "The ModelNode object of the child node index is not provided.");
--- a/src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc
+++ b/src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc
@@ -188,10 +188,12 @@ absl::Status BranchNodeImpl::Apply(LabelerEvent& event) const {
   } else if (matcher_) {
     // Select by condition.
     selected_index = matcher_->GetFirstMatch(event);
-    if (selected_index == -1) {
+    if (selected_index < 0) {
       return absl::InvalidArgumentError(
           "No condition matches the input event.");
     }
+  } else {
+    return absl::InternalError("No select options is set for the BranchNode.");
   }
   const ChildNodeRef& selected_node = child_nodes_[selected_index];
   if (selected_node.index() != 1) {
--- a/src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc
+++ b/src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc
@@ -29,6 +29,11 @@
 
 namespace wfa_virtual_people {
 
+enum SelectBranchBy {
+  CHANCE = 1,
+  CONDITION = 2,
+};
+
 // Converts @branch to a child node and appends to @child_nodes.
 absl::Status AppendChildNode(
     const BranchNode::Branch& branch,
@@ -69,27 +74,33 @@ absl::StatusOr<std::unique_ptr<BranchNodeImpl>> BranchNodeImpl::Build(
     RETURN_IF_ERROR(AppendChildNode(branch, factory, child_nodes));
   }
 
-  // Checks if there is any
-  //   @branch_node.branches.chance and
-  //   @branch_node.branches.condition
-  // Breaks if at least one @branch_node.branches has neither chance nor
-  // condition.
-  bool has_chance = false;
-  bool has_condition = false;
+  // If all @branch_node.branches have chance, use chance.
+  // If all @branch_node.branches have condition, use condition.
+  // Else return error.
+  SelectBranchBy select_by;
+  if (branch_node.branches(0).has_chance()) {
+    select_by = SelectBranchBy::CHANCE;
+  } else if (branch_node.branches(0).has_condition()) {
+    select_by = SelectBranchBy::CONDITION;
+  } else {
+    return absl::InvalidArgumentError(
+        "BranchNode must have one of chance and condition.");
+  }
+
   for (const BranchNode::Branch& branch : branch_node.branches()) {
-    if (branch.has_chance()) {
-      has_chance = true;
-    } else if (branch.has_condition()) {
-      has_condition = true;
-    } else {
+    if (select_by == SelectBranchBy::CHANCE && !branch.has_chance()) {
       return absl::InvalidArgumentError(
-          "BranchNode must have one of chance and condition.");
+          "BranchNode must have chance set.");
+    }
+    if (select_by == SelectBranchBy::CONDITION && !branch.has_condition()) {
+      return absl::InvalidArgumentError(
+          "BranchNode must have condition set.");
     }
   }
 
   std::unique_ptr<DistributedConsistentHashing> hashing = nullptr;
   std::unique_ptr<FieldFiltersMatcher> matcher = nullptr;
-  if (has_chance && !has_condition) {
+  if (select_by == SelectBranchBy::CHANCE) {
     // All branches have chance set. The child node is selected by cosistent
     // hashing.
     std::vector<DistributionChoice> distribution;
@@ -100,7 +111,8 @@ absl::StatusOr<std::unique_ptr<BranchNodeImpl>> BranchNodeImpl::Build(
     }
     ASSIGN_OR_RETURN(
         hashing, DistributedConsistentHashing::Build(std::move(distribution)));
-  } else if (!has_chance && has_condition) {
+  }
+  if (select_by == SelectBranchBy::CONDITION) {
     // All branches have condition set. The child node is selected by condition
     // matching.
     std::vector<const FieldFilterProto*> filter_configs;
@@ -108,9 +120,6 @@ absl::StatusOr<std::unique_ptr<BranchNodeImpl>> BranchNodeImpl::Build(
       filter_configs.emplace_back(&branch.condition());
     }
     ASSIGN_OR_RETURN(matcher, FieldFiltersMatcher::Build(filter_configs));
-  } else {
-    return absl::InvalidArgumentError(
-        "All branches must have same type of select_by set.");
   }
 
   return absl::make_unique<BranchNodeImpl>(
--- a/src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc
+++ b/src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc
@@ -114,21 +114,28 @@ absl::StatusOr<std::unique_ptr<BranchNodeImpl>> BranchNodeImpl::Build(
   for (const BranchNode::Branch& branch : branch_node.branches()) {
     if (select_by == SelectBranchBy::CHANCE && !branch.has_chance()) {
       return absl::InvalidArgumentError(
-          "BranchNode must have chance set.");
+          "All branches should use the same select_by type.");
     }
     if (select_by == SelectBranchBy::CONDITION && !branch.has_condition()) {
       return absl::InvalidArgumentError(
-          "BranchNode must have condition set.");
+          "All branches should use the same select_by type.");
     }
   }
 
   std::unique_ptr<DistributedConsistentHashing> hashing = nullptr;
   std::unique_ptr<FieldFiltersMatcher> matcher = nullptr;
-  if (select_by == SelectBranchBy::CHANCE) {
-    ASSIGN_OR_RETURN(hashing, BuildHashing(branch_node.branches()));
-  }
-  if (select_by == SelectBranchBy::CONDITION) {
-    ASSIGN_OR_RETURN(matcher, BuildMatcher(branch_node.branches()));
+  switch (select_by) {
+    case SelectBranchBy::CHANCE: {
+      ASSIGN_OR_RETURN(hashing, BuildHashing(branch_node.branches()));
+      break;
+    }
+    case SelectBranchBy::CONDITION: {
+      ASSIGN_OR_RETURN(matcher, BuildMatcher(branch_node.branches()));
+      break;
+    }
+    default:
+      // This should never happen.
+      return absl::InternalError("Invalid select_by type.");
   }
 
   return absl::make_unique<BranchNodeImpl>(
@@ -158,7 +165,7 @@ BranchNodeImpl::BranchNodeImpl(
 absl::Status ResolveChildReference(
     ChildNodeRef& child_node,
     absl::flat_hash_map<uint32_t, std::unique_ptr<ModelNode>>& node_refs) {
-  if (uint32_t* node_index = std::get_if<0>(&child_node)) {
+  if (uint32_t* node_index = std::get_if<uint32_t>(&child_node)) {
     // The child node is referenced by node index. Need to resolve to the
     // ModelNode object.
     // The owner of the corresponding ModelNode pointer will be its parent node.
@@ -171,7 +178,8 @@ absl::Status ResolveChildReference(
     child_node.emplace<1>(std::move(node.mapped()));
   }
 
-  if (std::unique_ptr<ModelNode>* node = std::get_if<1>(&child_node)) {
+  if (std::unique_ptr<ModelNode>* node =
+      std::get_if<std::unique_ptr<ModelNode>>(&child_node)) {
     // Resolve the child node references of the sub tree here, because this node
     // is the only owner of the pointers to the child nodes.
     return (*node)->ResolveChildReferences(node_refs);
@@ -189,7 +197,7 @@ absl::Status BranchNodeImpl::ResolveChildReferences(
 }
 
 absl::Status BranchNodeImpl::Apply(LabelerEvent& event) const {
-  int selected_index = -1;
+  int selected_index = kNoMatchingIndex;
   if (hashing_) {
     // Select by chance.
     selected_index = hashing_->Hash(
@@ -197,15 +205,17 @@ absl::Status BranchNodeImpl::Apply(LabelerEvent& event) const {
   } else if (matcher_) {
     // Select by condition.
     selected_index = matcher_->GetFirstMatch(event);
-    if (selected_index < 0) {
+    if (selected_index == kNoMatchingIndex) {
       return absl::InvalidArgumentError(
           "No condition matches the input event.");
     }
   } else {
     return absl::InternalError("No select options is set for the BranchNode.");
   }
+
   const ChildNodeRef& selected_node = child_nodes_[selected_index];
-  if (const std::unique_ptr<ModelNode>* node = std::get_if<1>(&selected_node)) {
+  if (const std::unique_ptr<ModelNode>* node =
+      std::get_if<std::unique_ptr<ModelNode>>(&selected_node)) {
     return (*node)->Apply(event);
   }
   return absl::FailedPreconditionError(
--- a/src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc
+++ b/src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc
@@ -39,7 +39,8 @@ absl::Status AppendChildNode(
     // object in ResolveChildReferences.
     child_nodes.emplace_back(branch.node_index());
     return absl::OkStatus();
-  } else if (branch.has_node()) {
+  }
+  if (branch.has_node()) {
     // Create the ModelNode object and store.
     child_nodes.emplace_back();
     ASSIGN_OR_RETURN(child_nodes.back(), factory.NewModelNode(branch.node()));
--- a/src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc
+++ b/src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc
@@ -196,11 +196,11 @@ absl::Status BranchNodeImpl::Apply(LabelerEvent& event) const {
     return absl::InternalError("No select options is set for the BranchNode.");
   }
   const ChildNodeRef& selected_node = child_nodes_[selected_index];
-  if (selected_node.index() != 1) {
-    return absl::FailedPreconditionError(
-        "The child node reference must be resolved before apply.");
+  if (const std::unique_ptr<ModelNode>* node = std::get_if<1>(&selected_node)) {
+    return (*node)->Apply(event);
   }
-  return std::get<1>(selected_node)->Apply(event);
+  return absl::FailedPreconditionError(
+      "The child node reference must be resolved before apply.");
 }
 
 }  // namespace wfa_virtual_people
--- a/src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc
+++ b/src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc
@@ -162,13 +162,13 @@ absl::Status ResolveChildReference(
     child_node.emplace<1>(std::move(node.mapped()));
   }
 
-  if (child_node.index() != 1) {
-    return absl::InternalError(
-        "Neither node index nor ModelNode object is set for a child node.");
+  if (std::unique_ptr<ModelNode>* node = std::get_if<1>(&child_node)) {
+    // Resolve the child node references of the sub tree here, because this node
+    // is the only owner of the pointers to the child nodes.
+    return (*node)->ResolveChildReferences(node_refs);
   }
-  // Resolve the child node references of the sub tree here, because this node
-  // is the only owner of the pointers to the child nodes.
-  return std::get<1>(child_node)->ResolveChildReferences(node_refs);
+  return absl::InternalError(
+      "Neither node index nor ModelNode object is set for a child node.");
 }
 
 absl::Status BranchNodeImpl::ResolveChildReferences(
--- a/src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc
+++ b/src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc
@@ -1,0 +1,194 @@
+// Copyright 2021 The Cross-Media Measurement Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "wfa/virtual_people/core/model/branch_node_impl.h"
+
+#include "absl/container/flat_hash_map.h"
+#include "absl/memory/memory.h"
+#include "absl/status/status.h"
+#include "absl/status/statusor.h"
+#include "absl/strings/str_cat.h"
+#include "absl/strings/string_view.h"
+#include "src/main/proto/wfa/virtual_people/common/model.pb.h"
+#include "wfa/measurement/common/macros.h"
+#include "wfa/virtual_people/core/model/model_node.h"
+#include "wfa/virtual_people/core/model/model_node_factory.h"
+#include "wfa/virtual_people/core/model/utils/distributed_consistent_hashing.h"
+#include "wfa/virtual_people/core/model/utils/field_filters_matcher.h"
+
+namespace wfa_virtual_people {
+
+// Converts @branch to a child node and appends to @child_nodes.
+absl::Status AppendChildNode(
+    const BranchNode::Branch& branch,
+    const ModelNodeFactory& factory,
+    std::vector<ChildNodeRef>* child_nodes) {
+  if (branch.has_node_index()) {
+    // Store the index of the child node. Will be resolved to the ModelNode
+    // object in ResolveChildReferences.
+    child_nodes->emplace_back(branch.node_index());
+    return absl::OkStatus();
+  } else if (branch.has_node()) {
+    // Create the ModelNode object and store.
+    child_nodes->emplace_back();
+    ASSIGN_OR_RETURN(child_nodes->back(), factory.NewModelNode(branch.node()));
+    return absl::OkStatus();
+  }
+  return absl::InvalidArgumentError(
+      "BranchNode must have one of node_index and node.");
+}
+
+absl::StatusOr<std::unique_ptr<BranchNodeImpl>> BranchNodeImpl::Build(
+    const CompiledNode& node_config) {
+  if (!node_config.has_branch_node()) {
+    return absl::InvalidArgumentError("This is not a branch node.");
+  }
+  const BranchNode& branch_node = node_config.branch_node();
+  if (branch_node.branches_size() == 0) {
+    return absl::InvalidArgumentError(
+        "BranchNode must have at least 1 branch.");
+  }
+
+  // Converts each Branch to ChildNodeRef. Breaks if encounters any error
+  // status.
+  std::vector<ChildNodeRef> child_nodes;
+  ModelNodeFactory factory;
+  for (const BranchNode::Branch& branch : branch_node.branches()) {
+    RETURN_IF_ERROR(AppendChildNode(branch, factory, &child_nodes));
+  }
+
+  // Checks if there is any
+  //   @branch_node.branches.chance and
+  //   @branch_node.branches.condition
+  // Breaks if at least one @branch_node.branches has neither chance nor
+  // condition.
+  bool has_chance = false;
+  bool has_condition = false;
+  for (const BranchNode::Branch& branch : branch_node.branches()) {
+    if (branch.has_chance()) {
+      has_chance = true;
+    } else if (branch.has_condition()) {
+      has_condition = true;
+    } else {
+      return absl::InvalidArgumentError(
+          "BranchNode must have one of chance and condition.");
+    }
+  }
+
+  std::unique_ptr<DistributedConsistentHashing> hashing = nullptr;
+  std::unique_ptr<FieldFiltersMatcher> matcher = nullptr;
+  if (has_chance && !has_condition) {
+    // All branches have chance set. The child node is selected by cosistent
+    // hashing.
+    std::vector<DistributionChoice> distribution;
+    int index = 0;
+    for (const BranchNode::Branch& branch : branch_node.branches()) {
+      distribution.emplace_back(DistributionChoice({index, branch.chance()}));
+      ++index;
+    }
+    ASSIGN_OR_RETURN(
+        hashing, DistributedConsistentHashing::Build(std::move(distribution)));
+  } else if (!has_chance && has_condition) {
+    // All branches have condition set. The child node is selected by condition
+    // matching.
+    std::vector<const FieldFilterProto*> filter_configs;
+    for (const BranchNode::Branch& branch : branch_node.branches()) {
+      filter_configs.emplace_back(&branch.condition());
+    }
+    ASSIGN_OR_RETURN(matcher, FieldFiltersMatcher::Build(filter_configs));
+  } else {
+    return absl::InvalidArgumentError(
+        "All branches must have same type of select_by set.");
+  }
+
+  return absl::make_unique<BranchNodeImpl>(
+      node_config,
+      std::move(child_nodes),
+      std::move(hashing),
+      branch_node.random_seed(),
+      std::move(matcher));
+}
+
+BranchNodeImpl::BranchNodeImpl(
+    const CompiledNode& node_config,
+    std::vector<ChildNodeRef>&& child_nodes,
+    std::unique_ptr<DistributedConsistentHashing> hashing,
+    absl::string_view random_seed,
+    std::unique_ptr<FieldFiltersMatcher> matcher):
+    ModelNode(node_config),
+    child_nodes_(std::move(child_nodes)),
+    hashing_(std::move(hashing)),
+    random_seed_(random_seed),
+    matcher_(std::move(matcher)) {}
+
+// If @child_node is set as an index, replaces it with the corresponding
+// ModelNode object if found in @node_refs, and deletes the index / ModelNode
+// pair from @node_refs.
+// Also resolve the child references of the sub-tree of the @child_node.
+absl::Status ResolveChildReference(
+    ChildNodeRef& child_node,
+    absl::flat_hash_map<uint32_t, std::unique_ptr<ModelNode>>* node_refs) {
+  if (child_node.index() == 0) {
+    // The child node is referenced by node index. Need to resolve to the
+    // ModelNode object.
+    uint32_t node_index = std::get<0>(child_node);
+    if (node_refs->find(node_index) == node_refs->end()) {
+      return absl::InvalidArgumentError(
+          "The ModelNode object of the child ndoe index is not provided.");
+    }
+    child_node.emplace<1>(std::move(node_refs->at(node_index)));
+    // The owner of the corresponding ModelNode pointer is its parent node
+    // now. Delete the key value pair.
+    node_refs->erase(node_index);
+  }
+
+  if (child_node.index() != 1) {
+    return absl::InternalError(
+        "Neither node index nor ModelNode object is set for a child node.");
+  }
+  // Resolve the child node references of the sub tree here, because this node
+  // is the only owner of the pointers to the child nodes.
+  return std::get<1>(child_node)->ResolveChildReferences(node_refs);
+}
+
+absl::Status BranchNodeImpl::ResolveChildReferences(
+    absl::flat_hash_map<uint32_t, std::unique_ptr<ModelNode>>* node_refs) {
+  for (ChildNodeRef& child_node : child_nodes_) {
+    RETURN_IF_ERROR(ResolveChildReference(child_node, node_refs));
+  }
+  return absl::OkStatus();
+}
+
+absl::Status BranchNodeImpl::Apply(LabelerEvent* event) const {
+  int selected_index = -1;
+  if (hashing_) {
+    // Select by chance.
+    selected_index = hashing_->Hash(
+        absl::StrCat(random_seed_, event->acting_fingerprint()));
+  } else if (matcher_) {
+    // Select by condition.
+    selected_index = matcher_->GetFirstMatch(*event);
+    if (selected_index == -1) {
+      return absl::InvalidArgumentError(
+          "No condition matches the input event.");
+    }
+  }
+  if (child_nodes_[selected_index].index() != 1) {
+    return absl::FailedPreconditionError(
+        "The child node reference must be resolved before apply.");
+  }
+  return std::get<1>(child_nodes_[selected_index])->Apply(event);
+}
+
+}  // namespace wfa_virtual_people
--- a/src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc
+++ b/src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc
@@ -101,7 +101,7 @@ absl::StatusOr<std::unique_ptr<BranchNodeImpl>> BranchNodeImpl::Build(
   std::unique_ptr<DistributedConsistentHashing> hashing = nullptr;
   std::unique_ptr<FieldFiltersMatcher> matcher = nullptr;
   if (select_by == SelectBranchBy::CHANCE) {
-    // All branches have chance set. The child node is selected by cosistent
+    // All branches have chance set. The child node is selected by consistent
     // hashing.
     std::vector<DistributionChoice> distribution;
     int index = 0;
--- a/src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc
+++ b/src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc
@@ -117,7 +117,7 @@ absl::StatusOr<std::unique_ptr<BranchNodeImpl>> BranchNodeImpl::Build(
       break;
     }
     default:
-      // This should never happen.
+      // No select_by is set.
       return absl::InvalidArgumentError(
           "BranchNode must have one of chance and condition.");
   }
--- a/src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc
+++ b/src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc
@@ -29,11 +29,6 @@
 
 namespace wfa_virtual_people {
 
-enum SelectBranchBy {
-  CHANCE = 1,
-  CONDITION = 2,
-};
-
 // Converts @branch to a child node and appends to @child_nodes.
 absl::Status AppendChildNode(
     const BranchNode::Branch& branch,
@@ -101,22 +96,10 @@ absl::StatusOr<std::unique_ptr<BranchNodeImpl>> BranchNodeImpl::Build(
   // If all @branch_node.branches have chance, use chance.
   // If all @branch_node.branches have condition, use condition.
   // Else return error.
-  SelectBranchBy select_by;
-  if (branch_node.branches(0).has_chance()) {
-    select_by = SelectBranchBy::CHANCE;
-  } else if (branch_node.branches(0).has_condition()) {
-    select_by = SelectBranchBy::CONDITION;
-  } else {
-    return absl::InvalidArgumentError(
-        "BranchNode must have one of chance and condition.");
-  }
-
+  BranchNode::Branch::SelectByCase select_by_case =
+      branch_node.branches(0).select_by_case();
   for (const BranchNode::Branch& branch : branch_node.branches()) {
-    if (select_by == SelectBranchBy::CHANCE && !branch.has_chance()) {
-      return absl::InvalidArgumentError(
-          "All branches should use the same select_by type.");
-    }
-    if (select_by == SelectBranchBy::CONDITION && !branch.has_condition()) {
+    if (select_by_case != branch.select_by_case()) {
       return absl::InvalidArgumentError(
           "All branches should use the same select_by type.");
     }
@@ -124,18 +107,19 @@ absl::StatusOr<std::unique_ptr<BranchNodeImpl>> BranchNodeImpl::Build(
 
   std::unique_ptr<DistributedConsistentHashing> hashing = nullptr;
   std::unique_ptr<FieldFiltersMatcher> matcher = nullptr;
-  switch (select_by) {
-    case SelectBranchBy::CHANCE: {
+  switch (select_by_case) {
+    case BranchNode::Branch::kChance: {
       ASSIGN_OR_RETURN(hashing, BuildHashing(branch_node.branches()));
       break;
     }
-    case SelectBranchBy::CONDITION: {
+    case BranchNode::Branch::kCondition: {
       ASSIGN_OR_RETURN(matcher, BuildMatcher(branch_node.branches()));
       break;
     }
     default:
       // This should never happen.
-      return absl::InternalError("Invalid select_by type.");
+      return absl::InvalidArgumentError(
+          "BranchNode must have one of chance and condition.");
   }
 
   return absl::make_unique<BranchNodeImpl>(
--- a/src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc
+++ b/src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc
@@ -144,14 +144,14 @@ absl::Status ResolveChildReference(
     // The child node is referenced by node index. Need to resolve to the
     // ModelNode object.
     uint32_t node_index = std::get<0>(child_node);
-    if (node_refs.find(node_index) == node_refs.end()) {
+    // The owner of the corresponding ModelNode pointer will be its parent node.
+    // Gets the key value pair, and deletes them from the map.
+    auto nh = node_refs.extract(node_index);
+    if (nh.empty()) {
       return absl::InvalidArgumentError(
-          "The ModelNode object of the child ndoe index is not provided.");
+          "The ModelNode object of the child node index is not provided.");
     }
-    child_node.emplace<1>(std::move(node_refs.at(node_index)));
-    // The owner of the corresponding ModelNode pointer is its parent node
-    // now. Delete the key value pair.
-    node_refs.erase(node_index);
+    child_node.emplace<1>(std::move(nh.mapped()));
   }
 
   if (child_node.index() != 1) {
@@ -185,11 +185,12 @@ absl::Status BranchNodeImpl::Apply(LabelerEvent& event) const {
           "No condition matches the input event.");
     }
   }
-  if (child_nodes_[selected_index].index() != 1) {
+  const ChildNodeRef& selected_node = child_nodes_[selected_index];
+  if (selected_node.index() != 1) {
     return absl::FailedPreconditionError(
         "The child node reference must be resolved before apply.");
   }
-  return std::get<1>(child_nodes_[selected_index])->Apply(event);
+  return std::get<1>(selected_node)->Apply(event);
 }
 
 }  // namespace wfa_virtual_people
--- a/src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc
+++ b/src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc
@@ -55,6 +55,30 @@ absl::Status AppendChildNode(
       "BranchNode must have one of node_index and node.");
 }
 
+// Builds hashing based on the chances distribution.
+// All branches must have chance set.
+absl::StatusOr<std::unique_ptr<DistributedConsistentHashing>> BuildHashing(
+    const google::protobuf::RepeatedPtrField<BranchNode::Branch>& branches) {
+  std::vector<DistributionChoice> distribution;
+  int index = 0;
+  for (const BranchNode::Branch& branch : branches) {
+    distribution.emplace_back(DistributionChoice({index, branch.chance()}));
+    ++index;
+  }
+  return DistributedConsistentHashing::Build(std::move(distribution));
+}
+
+// Builds matcher based on the conditions.
+// All branches must have condition set.
+absl::StatusOr<std::unique_ptr<FieldFiltersMatcher>> BuildMatcher(
+    const google::protobuf::RepeatedPtrField<BranchNode::Branch>& branches) {
+  std::vector<const FieldFilterProto*> filter_configs;
+  for (const BranchNode::Branch& branch : branches) {
+    filter_configs.emplace_back(&branch.condition());
+  }
+  return FieldFiltersMatcher::Build(filter_configs);
+}
+
 absl::StatusOr<std::unique_ptr<BranchNodeImpl>> BranchNodeImpl::Build(
     const CompiledNode& node_config) {
   if (!node_config.has_branch_node()) {
@@ -101,25 +125,10 @@ absl::StatusOr<std::unique_ptr<BranchNodeImpl>> BranchNodeImpl::Build(
   std::unique_ptr<DistributedConsistentHashing> hashing = nullptr;
   std::unique_ptr<FieldFiltersMatcher> matcher = nullptr;
   if (select_by == SelectBranchBy::CHANCE) {
-    // All branches have chance set. The child node is selected by consistent
-    // hashing.
-    std::vector<DistributionChoice> distribution;
-    int index = 0;
-    for (const BranchNode::Branch& branch : branch_node.branches()) {
-      distribution.emplace_back(DistributionChoice({index, branch.chance()}));
-      ++index;
-    }
-    ASSIGN_OR_RETURN(
-        hashing, DistributedConsistentHashing::Build(std::move(distribution)));
+    ASSIGN_OR_RETURN(hashing, BuildHashing(branch_node.branches()));
   }
   if (select_by == SelectBranchBy::CONDITION) {
-    // All branches have condition set. The child node is selected by condition
-    // matching.
-    std::vector<const FieldFilterProto*> filter_configs;
-    for (const BranchNode::Branch& branch : branch_node.branches()) {
-      filter_configs.emplace_back(&branch.condition());
-    }
-    ASSIGN_OR_RETURN(matcher, FieldFiltersMatcher::Build(filter_configs));
+    ASSIGN_OR_RETURN(matcher, BuildMatcher(branch_node.branches()));
   }
 
   return absl::make_unique<BranchNodeImpl>(
--- a/src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc
+++ b/src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc
@@ -33,16 +33,16 @@ namespace wfa_virtual_people {
 absl::Status AppendChildNode(
     const BranchNode::Branch& branch,
     const ModelNodeFactory& factory,
-    std::vector<ChildNodeRef>* child_nodes) {
+    std::vector<ChildNodeRef>& child_nodes) {
   if (branch.has_node_index()) {
     // Store the index of the child node. Will be resolved to the ModelNode
     // object in ResolveChildReferences.
-    child_nodes->emplace_back(branch.node_index());
+    child_nodes.emplace_back(branch.node_index());
     return absl::OkStatus();
   } else if (branch.has_node()) {
     // Create the ModelNode object and store.
-    child_nodes->emplace_back();
-    ASSIGN_OR_RETURN(child_nodes->back(), factory.NewModelNode(branch.node()));
+    child_nodes.emplace_back();
+    ASSIGN_OR_RETURN(child_nodes.back(), factory.NewModelNode(branch.node()));
     return absl::OkStatus();
   }
   return absl::InvalidArgumentError(
@@ -65,7 +65,7 @@ absl::StatusOr<std::unique_ptr<BranchNodeImpl>> BranchNodeImpl::Build(
   std::vector<ChildNodeRef> child_nodes;
   ModelNodeFactory factory;
   for (const BranchNode::Branch& branch : branch_node.branches()) {
-    RETURN_IF_ERROR(AppendChildNode(branch, factory, &child_nodes));
+    RETURN_IF_ERROR(AppendChildNode(branch, factory, child_nodes));
   }
 
   // Checks if there is any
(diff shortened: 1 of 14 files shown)


efoxepstein (changes requested):
> Reviewed 12 of 12 files at r1.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/6)* status: all files reviewed, 8 unresolved discussions (waiting on @EvgSkv, @kunyuchen, @lenjoy, @mpatel24, @my2457, @oliver-amzn, @tcsnfkx, and @wliue)

___
*[src/main/cc/wfa/virtual_people/core/model/branch_node_impl.h, line 29 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/6#-Mb8Wisw065ab3QgX40H:-Mb8Wisw065ab3QgX40I:b-mzddf7) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/412cf85ece31880dab06bc932ed7c1e2e7eee50c/src/main/cc/wfa/virtual_people/core/model/branch_node_impl.h#L29)):*
> ```objc
> namespace wfa_virtual_people {
> 
> typedef std::variant<uint32_t, std::unique_ptr<ModelNode>> ChildNodeRef;
> ```

1. Please use `using` instead of `typedef`.
2. Please use `absl::variant` instead of `std::variant`.
3. Please document what each case means.
___
*[src/main/cc/wfa/virtual_people/core/model/branch_node_impl.h, line 81 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/6#-Mb8XAZB9utY5_6VAgDy:-Mb8XAZB9utY5_6VAgDz:bvwrvn0) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/412cf85ece31880dab06bc932ed7c1e2e7eee50c/src/main/cc/wfa/virtual_people/core/model/branch_node_impl.h#L81)):*
> ```objc
>   // TODO(@tcsnfkx): Resolve the child references of the UpdateTree if exists.
>   absl::Status ResolveChildReferences(
>       absl::flat_hash_map<uint32_t, std::unique_ptr<ModelNode>>* node_refs
> ```

Please use a mutable reference unless you want this to be nullable.
___
*[src/main/cc/wfa/virtual_people/core/model/branch_node_impl.h, line 86 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/6#-Mb8XEmA4PihNGwjrHKa:-Mb8XEmB5KO5wQVl87tI:bvwrvn0) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/412cf85ece31880dab06bc932ed7c1e2e7eee50c/src/main/cc/wfa/virtual_people/core/model/branch_node_impl.h#L86)):*
> ```objc
>   // Uses @hashing_ or @matcher_ to select one of the ModelNodes in
>   // @child_nodes_, and apply the selected ModelNode.
>   absl::Status Apply(LabelerEvent* event) const override;
> ```

Please use a mutable reference unless you want this to be nullable.
___
*[src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc, line 36 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/6#-Mb8XZe3CrwVJY-l-yC6:-Mb8XZe3CrwVJY-l-yC7:bvwrvn0) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/412cf85ece31880dab06bc932ed7c1e2e7eee50c/src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc#L36)):*
> ```cpp
>     const BranchNode::Branch& branch,
>     const ModelNodeFactory& factory,
>     std::vector<ChildNodeRef>* child_nodes) {
> ```

Please use a mutable reference unless you want this to be nullable.
___
*[src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc, line 42 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/6#-Mb8XLmY-3YQFHuzsQLN:-Mb8XLmY-3YQFHuzsQLO:bmp3h9g) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/412cf85ece31880dab06bc932ed7c1e2e7eee50c/src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc#L42)):*
> ```cpp
>     child_nodes->emplace_back(branch.node_index());
>     return absl::OkStatus();
>   } else if (branch.has_node()) {
> ```

`else` not needed because `if`-branch always returns.
___
*[src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc, line 150 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/6#-Mb8YbBrCa1r6fg5DjhB:-Mb8YbBrCa1r6fg5DjhC:bxhwdad) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/412cf85ece31880dab06bc932ed7c1e2e7eee50c/src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc#L150)):*
> ```cpp
>           "The ModelNode object of the child ndoe index is not provided.");
>     }
>     child_node.emplace<1>(std::move(node_refs->at(node_index)));
> ```

This is doing an unnecessary triple lookup.

How about:

```
auto it = node_refs->find(node_index);
if (it == node_refs->end()) { ... }
child_node.emplace<1>(std::move(it->second));
node_refs->erase(it);
```

or even:

```
auto node = node_refs->extract(node_index);
if (node.key() != node_index) { ... }
child_node.emplace<1>(std::move(it->second));
```
___
*[src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc, line 187 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/6#-Mb8_8Ci4LzOvIhJUV5P:-Mb8_8Cm8ssw0vzRUPZG:bzau9gj) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/412cf85ece31880dab06bc932ed7c1e2e7eee50c/src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc#L187)):*
> ```cpp
>     }
>   }
>   if (child_nodes_[selected_index].index() != 1) {
> ```

You're doing a double-lookup here, how about:

```
SomeType child_node = child_nodes_[selected_index];
```
___
*[src/main/cc/wfa/virtual_people/core/model/utils/field_filters_matcher.cc, line 38 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/6#-Mb8_OkyDgtHVUqIEQ7N:-Mb8_OkyDgtHVUqIEQ7O:b-fgexx3) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/412cf85ece31880dab06bc932ed7c1e2e7eee50c/src/main/cc/wfa/virtual_people/core/model/utils/field_filters_matcher.cc#L38)):*
> ```cpp
>   // Converts each FieldFilterProto to FieldFilter. Breaks if encounters any
>   // error status.
>   std::vector<std::unique_ptr<FieldFilter>> filters;
> ```

Here and elsewhere: how big are these vectors going to be typically? Consider using absl::InlinedVector: https://github.com/abseil/abseil-cpp/blob/master/absl/container/inlined_vector.h -- this could save a lot of heap allocations.

tcsnfkx (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/6)* status: 6 of 14 files reviewed, 8 unresolved discussions (waiting on @efoxepstein, @EvgSkv, @kunyuchen, @lenjoy, @mpatel24, @my2457, @oliver-amzn, and @wliue)

___
*[src/main/cc/wfa/virtual_people/core/model/branch_node_impl.h, line 29 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/6#-Mb8Wisw065ab3QgX40H:-MbAQCIMBrTv4yZCWc5N:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/412cf85ece31880dab06bc932ed7c1e2e7eee50c/src/main/cc/wfa/virtual_people/core/model/branch_node_impl.h#L29)):*
<details><summary><i>Previously, efoxepstein (Eli Fox-Epstein) wrote…</i></summary><blockquote>

1. Please use `using` instead of `typedef`.
2. Please use `absl::variant` instead of `std::variant`.
3. Please document what each case means.
</blockquote></details>

Done.
___
*[src/main/cc/wfa/virtual_people/core/model/branch_node_impl.h, line 81 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/6#-Mb8XAZB9utY5_6VAgDy:-MbAQHVoEuPQ0OG_A-ft:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/412cf85ece31880dab06bc932ed7c1e2e7eee50c/src/main/cc/wfa/virtual_people/core/model/branch_node_impl.h#L81)):*
<details><summary><i>Previously, efoxepstein (Eli Fox-Epstein) wrote…</i></summary><blockquote>

Please use a mutable reference unless you want this to be nullable.
</blockquote></details>

Done.
___
*[src/main/cc/wfa/virtual_people/core/model/branch_node_impl.h, line 86 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/6#-Mb8XEmA4PihNGwjrHKa:-MbAQIZeAAU8zMg7RI0F:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/412cf85ece31880dab06bc932ed7c1e2e7eee50c/src/main/cc/wfa/virtual_people/core/model/branch_node_impl.h#L86)):*
<details><summary><i>Previously, efoxepstein (Eli Fox-Epstein) wrote…</i></summary><blockquote>

Please use a mutable reference unless you want this to be nullable.
</blockquote></details>

Done.
___
*[src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc, line 36 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/6#-Mb8XZe3CrwVJY-l-yC6:-MbAQJxG2jhAGGp4aSBZ:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/412cf85ece31880dab06bc932ed7c1e2e7eee50c/src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc#L36)):*
<details><summary><i>Previously, efoxepstein (Eli Fox-Epstein) wrote…</i></summary><blockquote>

Please use a mutable reference unless you want this to be nullable.
</blockquote></details>

Done.
___
*[src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc, line 42 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/6#-Mb8XLmY-3YQFHuzsQLN:-MbAQMAkA0K2nVGosFdq:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/412cf85ece31880dab06bc932ed7c1e2e7eee50c/src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc#L42)):*
<details><summary><i>Previously, efoxepstein (Eli Fox-Epstein) wrote…</i></summary><blockquote>

`else` not needed because `if`-branch always returns.
</blockquote></details>

Done.
___
*[src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc, line 150 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/6#-Mb8YbBrCa1r6fg5DjhB:-MbAQRIH-ZS78kQRzgzB:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/412cf85ece31880dab06bc932ed7c1e2e7eee50c/src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc#L150)):*
<details><summary><i>Previously, efoxepstein (Eli Fox-Epstein) wrote…</i></summary><blockquote>

This is doing an unnecessary triple lookup.

How about:

```
auto it = node_refs->find(node_index);
if (it == node_refs->end()) { ... }
child_node.emplace<1>(std::move(it->second));
node_refs->erase(it);
```

or even:

```
auto node = node_refs->extract(node_index);
if (node.key() != node_index) { ... }
child_node.emplace<1>(std::move(it->second));
```
</blockquote></details>

Done.
___
*[src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc, line 187 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/6#-Mb8_8Ci4LzOvIhJUV5P:-MbAQTeoDxWT3_vt62lG:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/412cf85ece31880dab06bc932ed7c1e2e7eee50c/src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc#L187)):*
<details><summary><i>Previously, efoxepstein (Eli Fox-Epstein) wrote…</i></summary><blockquote>

You're doing a double-lookup here, how about:

```
SomeType child_node = child_nodes_[selected_index];
```
</blockquote></details>

Done.
___
*[src/main/cc/wfa/virtual_people/core/model/utils/field_filters_matcher.cc, line 38 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/6#-Mb8_OkyDgtHVUqIEQ7N:-MbAQWIx59L367JO9yQI:byf9q9o) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/412cf85ece31880dab06bc932ed7c1e2e7eee50c/src/main/cc/wfa/virtual_people/core/model/utils/field_filters_matcher.cc#L38)):*
<details><summary><i>Previously, efoxepstein (Eli Fox-Epstein) wrote…</i></summary><blockquote>

Here and elsewhere: how big are these vectors going to be typically? Consider using absl::InlinedVector: https://github.com/abseil/abseil-cpp/blob/master/absl/container/inlined_vector.h -- this could save a lot of heap allocations.
</blockquote></details>

We expect the size of the vectors to be larger than 20 in most cases. I think it is unnecessary to use absl::InlinedVector. We can visit this again if the performance is not good.

efoxepstein (approved):
> Reviewed 8 of 8 files at r2.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/6)* status: all files reviewed, 3 unresolved discussions (waiting on @EvgSkv, @kunyuchen, @lenjoy, @mpatel24, @my2457, @oliver-amzn, @tcsnfkx, and @wliue)

___
*[src/main/cc/wfa/virtual_people/core/model/branch_node_impl.h, line 43 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/6#-MbCUx8L30YrBQQ0Q0O7:-MbCUx8L30YrBQQ0Q0O8:bjjgypf) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/14eca30e80a93f25fb72a89d4f5ab81c3bb3ac2f/src/main/cc/wfa/virtual_people/core/model/branch_node_impl.h#L43)):*
> ```objc
> // TODO(@tcsnfkx): Implement attributes updater.
> // TODO(@tcsnfkx): Implement multiplicity.
> class BranchNodeImpl : public ModelNode {
> ```

Why BranchNodeImpl instead of just BranchNode?
___
*[src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc, line 79 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/6#-MbCWbqP34oDeI56gp8e:-MbCWbqP34oDeI56gp8f:b-k8fq8k) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/14eca30e80a93f25fb72a89d4f5ab81c3bb3ac2f/src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc#L79)):*
> ```cpp
>   bool has_chance = false;
>   bool has_condition = false;
>   for (const BranchNode::Branch& branch : branch_node.branches()) {
> ```

I'm wondering if it would be more efficient to just check the first element here instead of looping over everything. Then in the cases below you can have something like:

```
    for (const BranchNode::Branch& branch : branch_node.branches()) {
      if (!branch.has_chance()) { return absl::InvalidArgumentError(...); }
      distribution.emplace_back(DistributionChoice({index, branch.chance()}));
      ++index;
    }
 ```

Alternatively, right before you declare `hashing = nullptr` below, you can do:

```
if (has_chance == has_condition) return absl::InvalidArgumentError(...);
```

That allows you to simplify your if-statements to just `if (has_chance)` and `else`
___
*[src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc, line 150 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/6#-MbCXgKe-YHzLfrW2jCt:-MbCXgKe-YHzLfrW2jCu:b-y65s3v) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/14eca30e80a93f25fb72a89d4f5ab81c3bb3ac2f/src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc#L150)):*
https://google.github.io/styleguide/cppguide.html#General_Naming_Rules

> Minimize the use of abbreviations that would likely be unknown to someone outside your project (especially acronyms and initialisms). .... As a rule of thumb, an abbreviation is probably OK if it's listed in Wikipedia. 

I'd just write "node" here.

File: src/main/cc/wfa/virtual_people/core/model/branch_node_impl.h [hunk 1]
```diff
@@ -0,0 +1,116 @@
...
+  //
+  // Returns error status if any of the following happens:
+  //   Any index in @child_nodes_ does not have an entry in @node_refs.
+  //   Neither index nor ModelNode is set in at least one entry in
+  //     @child_nodes_.
+  //   It returns error status when resolving the child references for any node
+  //     in the sub-tree.
+  //
+  // TODO(@tcsnfkx): Resolve the child references of the UpdateTree if exists.
+  absl::Status ResolveChildReferences(
```
wliue, line 85: Should this be on ModelNode interface? We expect to only do it once starting from root node, correct?

src/main/cc/wfa/virtual_people/core/model/branch_node_impl.h
```diff
@@ -0,0 +1,116 @@
...
+      absl::flat_hash_map<uint32_t, std::unique_ptr<ModelNode>>& node_refs
+  ) override;
+
+  // Uses @hashing_ or @matcher_ to select one of the ModelNodes in
+  // @child_nodes_, and apply the selected ModelNode.
+  absl::Status Apply(LabelerEvent& event) const override;
+
+  BranchNodeImpl(const BranchNodeImpl&) = delete;
```
wliue, line 93: Move up to after ctor/dtor?

src/main/cc/wfa/virtual_people/core/model/branch_node_impl.h
```diff
@@ -0,0 +1,116 @@
...
+  // corresponding ModelNode objects.
+  std::vector<ChildNodeRef> child_nodes_;
+
+  // If chance is set in each branches in @node_config, hashing_ is set, and
+  // used to select a child node with random_seed_ when Apply is called.
+  std::unique_ptr<DistributedConsistentHashing> hashing_;
+  const std::string random_seed_;
+
+  // If condition is set in each branches in @node_config, matcher_ is set, and
+  // used to select the first child node whose condition matches.
```
wliue, line 110: Could add " when Apply is called" to match hashing_.

src/main/cc/wfa/virtual_people/core/model/branch_node_impl_test.cc
```diff
@@ -0,0 +1,362 @@
...
+      population_node {
+        pools {
+          population_offset: 10
+          total_population: 1
+        }
+        random_seed: "TestPopulationNodeSeed1"
+      }
+  )pb", &population_node_config_1));
+  ASSERT_OK_AND_ASSIGN(
+      node_refs[2],
```
wliue, line 136: -> node_refs[population_node_config_1.index()]

src/main/cc/wfa/virtual_people/core/model/branch_node_impl_test.cc
```diff
@@ -0,0 +1,362 @@
...
+      population_node {
+        pools {
+          population_offset: 20
+          total_population: 1
+        }
+        random_seed: "TestPopulationNodeSeed2"
+      }
+  )pb", &population_node_config_2));
+  ASSERT_OK_AND_ASSIGN(
+      node_refs[3],
```
wliue, line 152: -> node_refs[population_node_config_2.index()]

src/main/cc/wfa/virtual_people/core/model/branch_node_impl_test.cc
```diff
@@ -0,0 +1,362 @@
...
+  )pb", &config));
+  ASSERT_OK_AND_ASSIGN(
+      std::unique_ptr<ModelNode> node,
+      ModelNodeFactory().NewModelNode(config));
+
+  absl::flat_hash_map<uint32_t, std::unique_ptr<ModelNode>> node_refs;
+  EXPECT_THAT(
+      node->ResolveChildReferences(node_refs),
+      StatusIs(absl::StatusCode::kInvalidArgument, ""));
+}
```
wliue, line 359: Add a test case for nested child reference, i.e. a node_index child also uses node_index.

src/main/cc/wfa/virtual_people/core/model/model_node.h
```diff
@@ -33,8 +34,20 @@ class ModelNode {
...
+  // This method is to replace the child node indexes with the initialized
+  // ModelNode objects.
+  //
+  // @node_refs is the mapping from indexes to the ModelNode objects.
+  //
+  // In the ModelNode constructor, if node_index is used to represent the child
+  // node, we keep the indexes of the child nodes rather than replacing with the
+  // actual ModelNode objects immediately. And this method is required to be
+  // called to resolve those indexes of the child nodes.
+  virtual absl::Status ResolveChildReferences(
```
wliue, line 46: Should this be on ModelNode, or just a factory function?

src/main/cc/wfa/virtual_people/core/model/utils/field_filters_matcher.cc
```diff
@@ -0,0 +1,60 @@
...
+
+int FieldFiltersMatcher::GetFirstMatch(const LabelerEvent& event) const {
+  int index = 0;
+  for (auto& filter : filters_) {
+    if (filter->IsMatch(event)) {
+      return index;
+    }
+    ++index;
+  }
+  return -1;
```
wliue, line 57: Use a constant to indicate no_match. And the caller can check using the constant, rather than being coupled with this.

src/main/cc/wfa/virtual_people/core/model/utils/field_filters_matcher.h
```diff
@@ -0,0 +1,57 @@
...
+  // Never call the constructor directly.
+  explicit FieldFiltersMatcher(
+      std::vector<std::unique_ptr<FieldFilter>>&& filters):
+      filters_(std::move(filters)) {}
+
+  FieldFiltersMatcher(const FieldFiltersMatcher&) = delete;
+  FieldFiltersMatcher& operator=(const FieldFiltersMatcher&) = delete;
+
+  // Returns the index of the first matching FieldFilter in @filters_.
+  // Returns -1 if no matching.
```
wliue, line 46: Use a constant for no_match.

tcsnfkx (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/6)* status: 9 of 14 files reviewed, 22 unresolved discussions (waiting on @efoxepstein, @EvgSkv, @kunyuchen, @lenjoy, @mpatel24, @my2457, @oliver-amzn, and @wliue)

___
*[src/main/cc/wfa/virtual_people/core/model/branch_node_impl.h, line 43 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/6#-MbCUx8L30YrBQQ0Q0O7:-MbE1BOmD97ehy3TLunY:b-pkqb8k) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/14eca30e80a93f25fb72a89d4f5ab81c3bb3ac2f/src/main/cc/wfa/virtual_people/core/model/branch_node_impl.h#L43)):*
<details><summary><i>Previously, efoxepstein (Eli Fox-Epstein) wrote…</i></summary><blockquote>

Why BranchNodeImpl instead of just BranchNode?
</blockquote></details>

BranchNode is already used by the protobuf in the same namespace.
___
*[src/main/cc/wfa/virtual_people/core/model/branch_node_impl.h, line 52 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/6#-Mb8WfNMSo-q6SvXs04b-r2-52:-MbFShwh1qtOYuWoShUG:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/14eca30e80a93f25fb72a89d4f5ab81c3bb3ac2f/src/main/cc/wfa/virtual_people/core/model/branch_node_impl.h#L52)):*
<details><summary><i>Previously, wliue (Wei Liu) wrote…</i></summary><blockquote>

Nit: the meaning of "neither ... nor ... in at least one" is ambiguous.
</blockquote></details>

Done.
___
*[src/main/cc/wfa/virtual_people/core/model/branch_node_impl.h, line 85 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/6#-Mb8WfNMSo-q6SvXs04b-r2-85:-MbEBddQ5ZodAxKyYAUG:b-j2ki7i) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/14eca30e80a93f25fb72a89d4f5ab81c3bb3ac2f/src/main/cc/wfa/virtual_people/core/model/branch_node_impl.h#L85)):*
<details><summary><i>Previously, wliue (Wei Liu) wrote…</i></summary><blockquote>

Should this be on ModelNode interface? We expect to only do it once starting from root node, correct?
</blockquote></details>

Yes. This PR added this method to ModelNode interface also. In Labeler API, we will call this on root node.
___
*[src/main/cc/wfa/virtual_people/core/model/branch_node_impl.h, line 93 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/6#-Mb8WfNMSo-q6SvXs04b-r2-93:-MbFSx9v5yMvh-1WXY45:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/14eca30e80a93f25fb72a89d4f5ab81c3bb3ac2f/src/main/cc/wfa/virtual_people/core/model/branch_node_impl.h#L93)):*
<details><summary><i>Previously, wliue (Wei Liu) wrote…</i></summary><blockquote>

Move up to after ctor/dtor?
</blockquote></details>

Done.
___
*[src/main/cc/wfa/virtual_people/core/model/branch_node_impl.h, line 110 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/6#-Mb8WfNMSo-q6SvXs04b-r2-110:-MbFT-bbC0QcsD-AHcmV:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/14eca30e80a93f25fb72a89d4f5ab81c3bb3ac2f/src/main/cc/wfa/virtual_people/core/model/branch_node_impl.h#L110)):*
<details><summary><i>Previously, wliue (Wei Liu) wrote…</i></summary><blockquote>

Could add " when Apply is called" to match hashing_.
</blockquote></details>

Done.
___
*[src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc, line 72 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/6#-Mb8WfNMSo-q6SvXs04a-r2-72:-MbFT5NJ73l2DtQOVexM:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/14eca30e80a93f25fb72a89d4f5ab81c3bb3ac2f/src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc#L72)):*
<details><summary><i>Previously, wliue (Wei Liu) wrote…</i></summary><blockquote>

chance and condition are oneof. Maybe the following is a little easier to reason about:
- if all have chance: use chance
- if all have condition: use condition
- else return error

Alternatively, use the first branch to determine whether to use chance or condition. Then check that the rest has the same setting.
</blockquote></details>

Done.
___
*[src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc, line 79 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/6#-MbCWbqP34oDeI56gp8e:-MbFT7VWAVniNWwN13C-:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/14eca30e80a93f25fb72a89d4f5ab81c3bb3ac2f/src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc#L79)):*
<details><summary><i>Previously, efoxepstein (Eli Fox-Epstein) wrote…</i></summary><blockquote>

I'm wondering if it would be more efficient to just check the first element here instead of looping over everything. Then in the cases below you can have something like:

```
    for (const BranchNode::Branch& branch : branch_node.branches()) {
      if (!branch.has_chance()) { return absl::InvalidArgumentError(...); }
      distribution.emplace_back(DistributionChoice({index, branch.chance()}));
      ++index;
    }
 ```

Alternatively, right before you declare `hashing = nullptr` below, you can do:

```
if (has_chance == has_condition) return absl::InvalidArgumentError(...);
```

That allows you to simplify your if-statements to just `if (has_chance)` and `else`
</blockquote></details>

Done.
___
*[src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc, line 93 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/6#-Mb8WfNMSo-q6SvXs04a-r2-93:-MbFTAb17AteCZDn0xCd:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/14eca30e80a93f25fb72a89d4f5ab81c3bb3ac2f/src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc#L93)):*
<details><summary><i>Previously, wliue (Wei Liu) wrote…</i></summary><blockquote>

Typo: consistent
</blockquote></details>

Done.
___
*[src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc, line 113 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/6#-Mb8WfNMSo-q6SvXs04a-r2-113:-MbFTCa-292H3KWyNH2C:bss4a2b) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/14eca30e80a93f25fb72a89d4f5ab81c3bb3ac2f/src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc#L113)):*
<details><summary><i>Previously, wliue (Wei Liu) wrote…</i></summary><blockquote>

-> have the same type
</blockquote></details>

Removed.
___
*[src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc, line 138 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/6#-Mb8WfNMSo-q6SvXs04a-r2-138:-MbFTJIX6BkTbHGeBeiK:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/14eca30e80a93f25fb72a89d4f5ab81c3bb3ac2f/src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc#L138)):*
<details><summary><i>Previously, wliue (Wei Liu) wrote…</i></summary><blockquote>

Could also mention that it is an error if not found.
</blockquote></details>

Done.
___
*[src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc, line 143 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/6#-Mb8WfNMSo-q6SvXs04a-r2-143:-MbFTKHH2sxcMFfwJWwJ:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/14eca30e80a93f25fb72a89d4f5ab81c3bb3ac2f/src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc#L143)):*
<details><summary><i>Previously, wliue (Wei Liu) wrote…</i></summary><blockquote>

Maybe use get_if<>?
</blockquote></details>

Done.
___
*[src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc, line 150 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/6#-MbCXgKe-YHzLfrW2jCt:-MbFTLER9dKq7-9oTQ7n:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/14eca30e80a93f25fb72a89d4f5ab81c3bb3ac2f/src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc#L150)):*
<details><summary><i>Previously, efoxepstein (Eli Fox-Epstein) wrote…</i></summary><blockquote>

https://google.github.io/styleguide/cppguide.html#General_Naming_Rules

> Minimize the use of abbreviations that would likely be unknown to someone outside your project (especially acronyms and initialisms). .... As a rule of thumb, an abbreviation is probably OK if it's listed in Wikipedia. 

I'd just write "node" here.
</blockquote></details>

Done.
___
*[src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc, line 157 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/6#-Mb8WfNMSo-q6SvXs04a-r2-157:-MbFTN69BsldRkelllHg:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/14eca30e80a93f25fb72a89d4f5ab81c3bb3ac2f/src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc#L157)):*
<details><summary><i>Previously, wliue (Wei Liu) wrote…</i></summary><blockquote>

Maybe use get_if<>?
</blockquote></details>

Done.
___
*[src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc, line 158 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/6#-Mb8WfNMSo-q6SvXs04a-r2-158:-MbEL0sC9ik1S07gS5qY:b-t0od5w) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/14eca30e80a93f25fb72a89d4f5ab81c3bb3ac2f/src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc#L158)):*
<details><summary><i>Previously, wliue (Wei Liu) wrote…</i></summary><blockquote>

Any particular reason to use InternalError? Maybe also use InvalidArgumentError to be consistent.
On a side note, there is a check in AppendChildNode, so this error would be caught before reaching here. 
</blockquote></details>

I use InternalError here because we expect it to be caught before reaching here, and if we get this error, it means some code above has bug rather than the input data is wrong.
___
*[src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc, line 180 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/6#-Mb8WfNMSo-q6SvXs04a-r2-180:-MbFTQuW8B-q0fOJZyXW:b-adpbcc) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/14eca30e80a93f25fb72a89d4f5ab81c3bb3ac2f/src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc#L180)):*
<details><summary><i>Previously, wliue (Wei Liu) wrote…</i></summary><blockquote>

Is it necessary to check matcher_? What if both hashing_ and matcher_ are null (which shouldn't happen)?
If check matcher_, then also need to check selected_index >= 0 after.
</blockquote></details>

Though it should be impossible to have both hashing_ and matcher_ unset, it is better to return error to catch bugs in future changes.
___
*[src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc, line 183 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/6#-Mb8WfNMSo-q6SvXs04a-r2-183:-MbFTt8b0pbS-RykUr05:b-jclqhc) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/14eca30e80a93f25fb72a89d4f5ab81c3bb3ac2f/src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc#L183)):*
<details><summary><i>Previously, wliue (Wei Liu) wrote…</i></summary><blockquote>

This is coupled with matcher_ implementation. Maybe use a constant there?
</blockquote></details>

Change to check for non-negative.
___
*[src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc, line 193 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/6#-Mb8WfNMSo-q6SvXs04a-r2-193:-MbFU-XM0XWBYrfRvLyE:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/14eca30e80a93f25fb72a89d4f5ab81c3bb3ac2f/src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc#L193)):*
<details><summary><i>Previously, wliue (Wei Liu) wrote…</i></summary><blockquote>

Maybe get_if<>?
</blockquote></details>

Done.
___
*[src/main/cc/wfa/virtual_people/core/model/branch_node_impl_test.cc, line 136 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/6#-Mb8WfNMSo-q6SvXs04c-r2-136:-MbFJUH22Yceh5QoHFsm:b-tc1iuu) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/14eca30e80a93f25fb72a89d4f5ab81c3bb3ac2f/src/main/cc/wfa/virtual_people/core/model/branch_node_impl_test.cc#L136)):*
<details><summary><i>Previously, wliue (Wei Liu) wrote…</i></summary><blockquote>

-> node_refs[population_node_config_1.index()]
</blockquote></details>

From the perspective of BranchNodeImpl, it does not matter if the key of the node is the same as the index. Here we just care about the key, and the index is ignored.
___
*[src/main/cc/wfa/virtual_people/core/model/branch_node_impl_test.cc, line 152 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/6#-Mb8WfNMSo-q6SvXs04c-r2-152:-MbFJo2kCsBXvNZ7pVbX:b2nreh6) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/14eca30e80a93f25fb72a89d4f5ab81c3bb3ac2f/src/main/cc/wfa/virtual_people/core/model/branch_node_impl_test.cc#L152)):*
<details><summary><i>Previously, wliue (Wei Liu) wrote…</i></summary><blockquote>

-> node_refs[population_node_config_2.index()]
</blockquote></details>

Same as above.
___
*[src/main/cc/wfa/virtual_people/core/model/branch_node_impl_test.cc, line 359 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/6#-Mb8WfNMSo-q6SvXs04c-r2-359:-MbFU2NYDC9K5q4L3nWF:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/14eca30e80a93f25fb72a89d4f5ab81c3bb3ac2f/src/main/cc/wfa/virtual_people/core/model/branch_node_impl_test.cc#L359)):*
<details><summary><i>Previously, wliue (Wei Liu) wrote…</i></summary><blockquote>

Add a test case for nested child reference, i.e. a node_index child also uses node_index.
</blockquote></details>

Done.
___
*[src/main/cc/wfa/virtual_people/core/model/model_node.h, line 46 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/6#-Mb8WfNMSo-q6SvXs04X-r2-46:-MbFMTUaEYkwy5K30NJA:b-909o1i) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/14eca30e80a93f25fb72a89d4f5ab81c3bb3ac2f/src/main/cc/wfa/virtual_people/core/model/model_node.h#L46)):*
<details><summary><i>Previously, wliue (Wei Liu) wrote…</i></summary><blockquote>

Should this be on ModelNode, or just a factory function?
</blockquote></details>

A factory function still requires to call the method in this interface to resolve the tree structure.
___
*[src/main/cc/wfa/virtual_people/core/model/utils/field_filters_matcher.h, line 46 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/6#-Mb8WfNNRQPk_JBmSa3C-r2-46:-MbFU4g4D236tOGUBwGG:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/14eca30e80a93f25fb72a89d4f5ab81c3bb3ac2f/src/main/cc/wfa/virtual_people/core/model/utils/field_filters_matcher.h#L46)):*
<details><summary><i>Previously, wliue (Wei Liu) wrote…</i></summary><blockquote>

Use a constant for no_match.
</blockquote></details>

Done.
___
*[src/main/cc/wfa/virtual_people/core/model/utils/field_filters_matcher.cc, line 57 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/6#-Mb8WfNMSo-q6SvXs04d-r2-57:-MbFU5sUCNG2ScTow1s5:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/14eca30e80a93f25fb72a89d4f5ab81c3bb3ac2f/src/main/cc/wfa/virtual_people/core/model/utils/field_filters_matcher.cc#L57)):*
<details><summary><i>Previously, wliue (Wei Liu) wrote…</i></summary><blockquote>

Use a constant to indicate no_match. And the caller can check using the constant, rather than being coupled with this.
</blockquote></details>

Done.

**Review by efoxepstein — APPROVED** [review]
> Reviewed 5 of 5 files at r3.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/6)* status: all files reviewed, 20 unresolved discussions (waiting on @EvgSkv, @kunyuchen, @lenjoy, @mpatel24, @my2457, @oliver-amzn, @tcsnfkx, and @wliue)

___
*[src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc, line 103 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/6#-MbHmqku4x8P7ZsNGbet:-MbHmqku4x8P7ZsNGbeu:bln9ezr) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/5a59cb38959d891dbbc82176f4c0fa0a46bd3a2d/src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc#L103)):*
> ```cpp
>   std::unique_ptr<DistributedConsistentHashing> hashing = nullptr;
>   std::unique_ptr<FieldFiltersMatcher> matcher = nullptr;
>   if (select_by == SelectBranchBy::CHANCE) {
> ```

Optional: consider pulling the bodies of these if-statements out into helper functions, so here you can have:

```
switch (select_by) {
  case CHANCE: ASSIGN_OR_RETURN(hashing, BuildDistributedConsistentHashing(branch_node)); break;
  case CONDITION: ASSIGN_OR_RETURN(matcher, BuildMatcher(branch_node)); break;
}
```

with helpers like:

```
absl::StatusOr<std::unique_ptr<DistributedConsistentHashing>> BuildDistributedConsistentHashing(const BranchNode& branch_node) {
    std::vector<DistributionChoice> distribution;
    int index = 0;
    for (const BranchNode::Branch& branch : branch_node.branches()) {
      distribution.emplace_back(DistributionChoice({index, branch.chance()}));
      ++index;
    }
    return DistributedConsistentHashing::Build(std::move(distribution));
}
```

**Review by tcsnfkx — COMMENTED** [review]
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/6)* status: 13 of 14 files reviewed, 19 unresolved discussions (waiting on @efoxepstein, @EvgSkv, @kunyuchen, @lenjoy, @mpatel24, @my2457, @oliver-amzn, and @wliue)

___
*[src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc, line 103 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/6#-MbHmqku4x8P7ZsNGbet:-MbI72Je2bQhCahWMOQW:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/5a59cb38959d891dbbc82176f4c0fa0a46bd3a2d/src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc#L103)):*
<details><summary><i>Previously, efoxepstein (Eli Fox-Epstein) wrote…</i></summary><blockquote>

Optional: consider pulling the bodies of these if-statements out into helper functions, so here you can have:

```
switch (select_by) {
  case CHANCE: ASSIGN_OR_RETURN(hashing, BuildDistributedConsistentHashing(branch_node)); break;
  case CONDITION: ASSIGN_OR_RETURN(matcher, BuildMatcher(branch_node)); break;
}
```

with helpers like:

```
absl::StatusOr<std::unique_ptr<DistributedConsistentHashing>> BuildDistributedConsistentHashing(const BranchNode& branch_node) {
    std::vector<DistributionChoice> distribution;
    int index = 0;
    for (const BranchNode::Branch& branch : branch_node.branches()) {
      distribution.emplace_back(DistributionChoice({index, branch.chance()}));
      ++index;
    }
    return DistributedConsistentHashing::Build(std::move(distribution));
}
```
</blockquote></details>

Done.

**Review by efoxepstein — CHANGES_REQUESTED** [review]
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/6)* status: 13 of 14 files reviewed, 20 unresolved discussions (waiting on @efoxepstein, @EvgSkv, @kunyuchen, @lenjoy, @mpatel24, @my2457, @oliver-amzn, @tcsnfkx, and @wliue)

___
*[src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc, line 128 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/6#-MbIAv_H4MWt4HP4Ybx9:-MbIAv_H4MWt4HP4YbxA:b-eqecky) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/26b3059faa2dd78b3e53aa6dcc707dfa798ff13e/src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc#L128)):*
> ```cpp
>   std::unique_ptr<FieldFiltersMatcher> matcher = nullptr;
>   if (select_by == SelectBranchBy::CHANCE) {
>     ASSIGN_OR_RETURN(hashing, BuildHashing(branch_node.branches()));
> ```

Please use a switch on `select_by` so we know it's exhaustive coverage.

**Review by efoxepstein — COMMENTED** [review]
> Reviewed 1 of 1 files at r4.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/6)* status: all files reviewed, 20 unresolved discussions (waiting on @EvgSkv, @kunyuchen, @lenjoy, @mpatel24, @my2457, @oliver-amzn, @tcsnfkx, and @wliue)

**Review by wliue — CHANGES_REQUESTED** [review]
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/6)* status: all files reviewed, 24 unresolved discussions (waiting on @EvgSkv, @kunyuchen, @lenjoy, @mpatel24, @my2457, @oliver-amzn, @tcsnfkx, and @wliue)

___
*[src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc, line 143 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/6#-Mb8WfNMSo-q6SvXs04a-r2-143:-MbJ-S7-A50Hu7lXuTik:b21k23j) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/14eca30e80a93f25fb72a89d4f5ab81c3bb3ac2f/src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc#L143)):*
<details><summary><i>Previously, tcsnfkx wrote…</i></summary><blockquote>

Done.
</blockquote></details>


Maybe absl::get_if by type?
___
*[src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc, line 157 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/6#-Mb8WfNMSo-q6SvXs04a-r2-157:-MbJ-LOk6kV0uIAS_ayE:b21k23j) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/14eca30e80a93f25fb72a89d4f5ab81c3bb3ac2f/src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc#L157)):*
<details><summary><i>Previously, tcsnfkx wrote…</i></summary><blockquote>

Done.
</blockquote></details>


Maybe absl::get_if by type?
___
*[src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc, line 183 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/6#-Mb8WfNMSo-q6SvXs04a-r2-183:-MbJ-DQzAmGr6-GhseYI:b-xcqjlp) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/14eca30e80a93f25fb72a89d4f5ab81c3bb3ac2f/src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc#L183)):*
<details><summary><i>Previously, tcsnfkx wrote…</i></summary><blockquote>

Change to check for non-negative.
</blockquote></details>


Check against the NoMatch constant.
___
*[src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc, line 193 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/6#-Mb8WfNMSo-q6SvXs04a-r2-193:-MbJ-oZP9AjTvY5OaguQ:b7xhmwv) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/14eca30e80a93f25fb72a89d4f5ab81c3bb3ac2f/src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc#L193)):*
<details><summary><i>Previously, tcsnfkx wrote…</i></summary><blockquote>

Done.
</blockquote></details>

Maybe absl::get_if by type?
___
*[src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc, line 117 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/6#-MbIw87D9huTmaQEmyCe:-MbIw87D9huTmaQEmyCf:b3cqbxp) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/26b3059faa2dd78b3e53aa6dcc707dfa798ff13e/src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc#L117)):*
> ```
> "BranchNode must have chance set."
> ```
This error message is not quite right. The error is more about "All branches should use the same select_by type."
___
*[src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc, line 121 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/6#-MbIwbxR6qaLhPOliH1d:-MbIwbxSBMBz153RHPh3:b1atelm) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/26b3059faa2dd78b3e53aa6dcc707dfa798ff13e/src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc#L121)):*
> ```
>           "BranchNode must have condition set.");
> ```
This error message is not quite right. The error is more about "All branches should use the same select_by type."
___
*[src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc, line 207 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/6#-MbJ6Wfp9e0f29CD-9Zf:-MbJ6Wfp9e0f29CD-9Zg:b-ovvdzc) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/26b3059faa2dd78b3e53aa6dcc707dfa798ff13e/src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc#L207)):*
> ```
>  const ChildNodeRef& selected_node = child_nodes_[selected_index];
> ```

Nit: add a blank line before.
___
*[src/main/cc/wfa/virtual_people/core/model/utils/field_filters_matcher.h, line 25 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/6#-MbJ5nU91Dg23J4HVfxh:-MbJ5nU91Dg23J4HVfxi:b-e4e3pm) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/26b3059faa2dd78b3e53aa6dcc707dfa798ff13e/src/main/cc/wfa/virtual_people/core/model/utils/field_filters_matcher.h#L25)):*
> ```
> constexpr int kNoMatchingIndex = -1;
> ```
inline this.

**Review by tcsnfkx — COMMENTED** [review]
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/6)* status: 13 of 14 files reviewed, 23 unresolved discussions (waiting on @efoxepstein, @EvgSkv, @kunyuchen, @lenjoy, @mpatel24, @my2457, @oliver-amzn, and @wliue)

___
*[src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc, line 143 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/6#-Mb8WfNMSo-q6SvXs04a-r2-143:-MbJWAT6DHpvgnMZFrb9:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/14eca30e80a93f25fb72a89d4f5ab81c3bb3ac2f/src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc#L143)):*
<details><summary><i>Previously, wliue (Wei Liu) wrote…</i></summary><blockquote>

Maybe absl::get_if by type?
</blockquote></details>

Done.
___
*[src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc, line 157 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/6#-Mb8WfNMSo-q6SvXs04a-r2-157:-MbJWAwJ606S3aSK2i9w:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/14eca30e80a93f25fb72a89d4f5ab81c3bb3ac2f/src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc#L157)):*
<details><summary><i>Previously, wliue (Wei Liu) wrote…</i></summary><blockquote>

Maybe absl::get_if by type?
</blockquote></details>

Done.
___
*[src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc, line 183 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/6#-Mb8WfNMSo-q6SvXs04a-r2-183:-MbJWD2-EWnFv_knJXKM:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/14eca30e80a93f25fb72a89d4f5ab81c3bb3ac2f/src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc#L183)):*
<details><summary><i>Previously, wliue (Wei Liu) wrote…</i></summary><blockquote>

Check against the NoMatch constant.
</blockquote></details>

Done.
___
*[src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc, line 193 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/6#-Mb8WfNMSo-q6SvXs04a-r2-193:-MbJWEd-7iUVrzrKdsDF:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/14eca30e80a93f25fb72a89d4f5ab81c3bb3ac2f/src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc#L193)):*
<details><summary><i>Previously, wliue (Wei Liu) wrote…</i></summary><blockquote>

Maybe absl::get_if by type?
</blockquote></details>

Done.
___
*[src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc, line 117 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/6#-MbIw87D9huTmaQEmyCe:-MbJW7Qk1-g7UHwbeFH0:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/26b3059faa2dd78b3e53aa6dcc707dfa798ff13e/src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc#L117)):*
<details><summary><i>Previously, wliue (Wei Liu) wrote…</i></summary><blockquote>

> ```
> "BranchNode must have chance set."
> ```
This error message is not quite right. The error is more about "All branches should use the same select_by type."
</blockquote></details>

Done.
___
*[src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc, line 121 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/6#-MbIwbxR6qaLhPOliH1d:-MbJW7zl9anrgazAMHWV:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/26b3059faa2dd78b3e53aa6dcc707dfa798ff13e/src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc#L121)):*
<details><summary><i>Previously, wliue (Wei Liu) wrote…</i></summary><blockquote>

> ```
>           "BranchNode must have condition set.");
> ```
This error message is not quite right. The error is more about "All branches should use the same select_by type."
</blockquote></details>

Done.
___
*[src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc, line 128 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/6#-MbIAv_H4MWt4HP4Ybx9:-MbJW9EQDwcitbbX8V9Q:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/26b3059faa2dd78b3e53aa6dcc707dfa798ff13e/src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc#L128)):*
<details><summary><i>Previously, efoxepstein (Eli Fox-Epstein) wrote…</i></summary><blockquote>

Please use a switch on `select_by` so we know it's exhaustive coverage.
</blockquote></details>

Done.
___
*[src/main/cc/wfa/virtual_people/core/model/utils/field_filters_matcher.h, line 25 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/6#-MbJ5nU91Dg23J4HVfxh:-MbJVaFI7NDkMg7u_YK4:bj5tk0g) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/26b3059faa2dd78b3e53aa6dcc707dfa798ff13e/src/main/cc/wfa/virtual_people/core/model/utils/field_filters_matcher.h#L25)):*
<details><summary><i>Previously, wliue (Wei Liu) wrote…</i></summary><blockquote>

> ```
> constexpr int kNoMatchingIndex = -1;
> ```
inline this.
</blockquote></details>

constexpr is implicitly inline.

**Review by wliue — CHANGES_REQUESTED** [review]
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/6)* status: 13 of 14 files reviewed, 23 unresolved discussions (waiting on @efoxepstein, @EvgSkv, @kunyuchen, @lenjoy, @mpatel24, @my2457, @oliver-amzn, @tcsnfkx, and @wliue)

___
*[src/main/cc/wfa/virtual_people/core/model/utils/field_filters_matcher.h, line 25 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/6#-MbJ5nU91Dg23J4HVfxh:-MbJgJjk4oT5VbZ6gjni:b-f56bgy) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/26b3059faa2dd78b3e53aa6dcc707dfa798ff13e/src/main/cc/wfa/virtual_people/core/model/utils/field_filters_matcher.h#L25)):*
<details><summary><i>Previously, tcsnfkx wrote…</i></summary><blockquote>

constexpr is implicitly inline.
</blockquote></details>

constexpr function and static member variable are implicitly inline, but not namespace-scope variable. 
See https://en.cppreference.com/w/cpp/language/inline

**Review by tcsnfkx — COMMENTED** [review]
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/6)* status: 12 of 14 files reviewed, 23 unresolved discussions (waiting on @efoxepstein, @EvgSkv, @kunyuchen, @lenjoy, @mpatel24, @my2457, @oliver-amzn, and @wliue)

___
*[src/main/cc/wfa/virtual_people/core/model/utils/field_filters_matcher.h, line 25 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/6#-MbJ5nU91Dg23J4HVfxh:-MbKoTy90iwKHpftWZ10:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/26b3059faa2dd78b3e53aa6dcc707dfa798ff13e/src/main/cc/wfa/virtual_people/core/model/utils/field_filters_matcher.h#L25)):*
<details><summary><i>Previously, wliue (Wei Liu) wrote…</i></summary><blockquote>

constexpr function and static member variable are implicitly inline, but not namespace-scope variable. 
See https://en.cppreference.com/w/cpp/language/inline 
</blockquote></details>

Done.

**Review by efoxepstein — CHANGES_REQUESTED** [review]
> Reviewed 1 of 1 files at r5, 1 of 1 files at r6, 1 of 1 files at r7.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/6)* status: all files reviewed, 23 unresolved discussions (waiting on @EvgSkv, @kunyuchen, @lenjoy, @mpatel24, @my2457, @oliver-amzn, @tcsnfkx, and @wliue)

___
*[src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc, line 34 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/6#-MbNM57BEoXOb-bcjAkS:-MbNM57BEoXOb-bcjAkT:bcberww) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/297d8b757cc779ed6c06964fc3d3549611494ff1/src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc#L34)):*
> ```cpp
> enum SelectBranchBy {
>   CHANCE = 1,
>   CONDITION = 2,
> ```

The proto compiler automatically generates an enum like this, accessible via the `oneof_name_case()` accessor: https://developers.google.com/protocol-buffers/docs/reference/cpp-generated#oneof

Maybe using that would be simpler because you can say `SelectByCase case = branch_node.branches(0).select_by_case();` and then later do:

```
if (branch.select_by_case() != case) { return an error; }
```
___
*[src/main/cc/wfa/virtual_people/core/model/utils/field_filters_matcher.h, line 25 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/6#-MbJ5nU91Dg23J4HVfxh:-MbNLJIx0TsBBBxtVw97:btmn6va) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/26b3059faa2dd78b3e53aa6dcc707dfa798ff13e/src/main/cc/wfa/virtual_people/core/model/utils/field_filters_matcher.h#L25)):*
<details><summary><i>Previously, tcsnfkx wrote…</i></summary><blockquote>

Done.
</blockquote></details>

For future reference: https://abseil.io/tips/140

**Review by wliue — APPROVED** [review]
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/6)* status: all files reviewed, 23 unresolved discussions (waiting on @EvgSkv, @kunyuchen, @lenjoy, @mpatel24, @my2457, @oliver-amzn, @tcsnfkx, and @wliue)

___
*[src/main/cc/wfa/virtual_people/core/model/model_node.h, line 46 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/6#-Mb8WfNMSo-q6SvXs04X-r2-46:-MbNdIpF0l8c2ZfD_3gZ:b-44ugve) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/14eca30e80a93f25fb72a89d4f5ab81c3bb3ac2f/src/main/cc/wfa/virtual_people/core/model/model_node.h#L46)):*
<details><summary><i>Previously, tcsnfkx wrote…</i></summary><blockquote>

A factory function still requires to call the method in this interface to resolve the tree structure.
</blockquote></details>

Could add a comment that this method is intended to be only used (by Labeler) during model loading.

**Review by tcsnfkx — COMMENTED** [review]
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/6)* status: 12 of 14 files reviewed, 23 unresolved discussions (waiting on @efoxepstein, @EvgSkv, @kunyuchen, @lenjoy, @mpatel24, @my2457, @oliver-amzn, and @wliue)

___
*[src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc, line 34 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/6#-MbNM57BEoXOb-bcjAkS:-MbNlwWO0Px1xuj5Ou8R:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/297d8b757cc779ed6c06964fc3d3549611494ff1/src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc#L34)):*
<details><summary><i>Previously, efoxepstein (Eli Fox-Epstein) wrote…</i></summary><blockquote>

The proto compiler automatically generates an enum like this, accessible via the `oneof_name_case()` accessor: https://developers.google.com/protocol-buffers/docs/reference/cpp-generated#oneof

Maybe using that would be simpler because you can say `SelectByCase case = branch_node.branches(0).select_by_case();` and then later do:

```
if (branch.select_by_case() != case) { return an error; }
```
</blockquote></details>

Done.
___
*[src/main/cc/wfa/virtual_people/core/model/model_node.h, line 46 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/6#-Mb8WfNMSo-q6SvXs04X-r2-46:-MbNlqvv2gPvXhS1rIrr:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/14eca30e80a93f25fb72a89d4f5ab81c3bb3ac2f/src/main/cc/wfa/virtual_people/core/model/model_node.h#L46)):*
<details><summary><i>Previously, wliue (Wei Liu) wrote…</i></summary><blockquote>

Could add a comment that this method is intended to be only used (by Labeler) during model loading. 
</blockquote></details>

Done.
___
*[src/main/cc/wfa/virtual_people/core/model/utils/field_filters_matcher.h, line 25 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/6#-MbJ5nU91Dg23J4HVfxh:-MbNlrrA-P46ITGKQUFj:b8zzliv) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/26b3059faa2dd78b3e53aa6dcc707dfa798ff13e/src/main/cc/wfa/virtual_people/core/model/utils/field_filters_matcher.h#L25)):*
<details><summary><i>Previously, efoxepstein (Eli Fox-Epstein) wrote…</i></summary><blockquote>

For future reference: https://abseil.io/tips/140
</blockquote></details>

Acknowledge.

**Review by efoxepstein — APPROVED** [review]
> Reviewed 2 of 2 files at r8.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/6)* status: all files reviewed, 22 unresolved discussions (waiting on @EvgSkv, @kunyuchen, @lenjoy, @mpatel24, @my2457, @oliver-amzn, and @wliue)

**Review by mpatel24 — APPROVED** [review]

File: src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc [hunk 10]
```diff
@@ -101,41 +96,30 @@ absl::StatusOr<std::unique_ptr<BranchNodeImpl>> BranchNodeImpl::Build(
...
       ASSIGN_OR_RETURN(hashing, BuildHashing(branch_node.branches()));
       break;
     }
-    case SelectBranchBy::CONDITION: {
+    case BranchNode::Branch::kCondition: {
       ASSIGN_OR_RETURN(matcher, BuildMatcher(branch_node.branches()));
       break;
     }
     default:
       // This should never happen.
```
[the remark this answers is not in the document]
tcsnfkx: Done.

PR #38, opened 2022-05-24
add jvm labeler

stevenwarejones:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/38)


SanjayVas (changes requested):
> Reviewed 1 of 38 files at r1, 9 of 17 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/38)* status: 10 of 40 files reviewed, 2 unresolved discussions (waiting on @robinsons, @stevenwarejones, @tcsnfkx, and @wliue)

*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/38#-N2sQ0Er6gZXRvdGq0lp:-N2sQ0Er6gZXRvdGq0lq:bv2qhkx) (no related file):*
I'm not as familiar with this codebase, but it looks like it's meant to be an "independent" library like [any-sketch](https://github.com/world-federation-of-advertisers/any-sketch).

If you want to add a Java version, it should probably be done similarly to [any-sketch-java](https://github.com/world-federation-of-advertisers/any-sketch-java). It should not add a significant number of dependencies that it doesn't already have (e.g. don't depend on common-jvm). The final library should be callable from Java (probably want to avoid using Kotlin unless the code is complex).

___
*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/38#-N2sUBNbBkw_TzooTFBn:-N2sUBNbBkw_TzooTFBo:bl4ppb8) (no related file):*
It looks like this is using JNI.  We have a high threshold for that, e.g. if we depend on a library that's only available in C++. We don't use JNI just to avoid having to reimplement our own code in a JVM language.

___

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/38)* status: 10 of 40 files reviewed, 2 unresolved discussions (waiting on @robinsons, @stevenwarejones, @tcsnfkx, and @wliue)

*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/38#-N2sUBNbBkw_TzooTFBn:-N2sVQkWAdezZcAhn_rC:bgwuugx) (no related file):*
<details><summary><i>Previously, SanjayVas (Sanjay Vasandani) wrote…</i></summary><blockquote>

It looks like this is using JNI.  We have a high threshold for that, e.g. if we depend on a library that's only available in C++. We don't use JNI just to avoid having to reimplement our own code in a JVM language.
</blockquote></details>

We should probably have a broader discussion about the context of this change w.r.t. any Halo feature requests involving implementing/maintaining a Java port of this library.

___

src/main/proto/wfa/virtual_people/core/labeler/input_labeler.proto
```diff
@@ -0,0 +1,35 @@
...
+option java_package = "org.wfanet.virtualpeople.core.labeler";
+option java_multiple_files = true;
+
+import "wfa/virtual_people/common/event.proto";
+import "wfa/virtual_people/common/label.proto";
+import "wfa/virtual_people/common/model.proto";
+
+// Request to Label Events
+message LabelEventsRequest {
+  CompiledNode root_node = 1;  // TODO: change this to repeated nodes
```
wangyaopw, line 28: I heard that a model file can be as big as several giga bytes. Not show how big the resulted compiledNode proto object can be. In other words, not sure if it is a good idea to call the label via JNI. Even if using JNI, we may want to share the memory between JVM and C++

wangyaopw:
> > _[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/38)_ status: 10 of 40 files reviewed, 2 unresolved discussions (waiting on @robinsons, @stevenwarejones, @tcsnfkx, and @wliue)
> 
> _[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/38#-N2sUBNbBkw_TzooTFBn:-N2sVQkWAdezZcAhn_rC:bgwuugx) (no related file):_
> 
> Previously, SanjayVas (Sanjay Vasandani) wrote…
> We should probably have a broader discussion about the context of this change w.r.t. any Halo feature requests involving implementing/maintaining a Java port of this library.

A java implementation of the labeler is definitely desirable, similar to the reason why we added the any-sketch java library. Since all EDPs are expected to use the any-sketch an VID labeler library.

The question is whether we should implement the library in Java directly, or call the c++ version via JNI. I would vote for the former, but it is more time consuming.

src/main/proto/wfa/virtual_people/core/labeler/input_labeler.proto
stevenwarejones: I agree on this one 100%

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/38)* status: 10 of 40 files reviewed, 3 unresolved discussions (waiting on @robinsons, @SanjayVas, @stevenwarejones, @tcsnfkx, @wangyaopw, and @wliue)

*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/38#-N2sUBNbBkw_TzooTFBn:-N3_9QG34uJ7UwPp6bXG:b1lwsg6) (no related file):*
<details><summary><i>Previously, wangyaopw (Yao Wang) wrote…</i></summary><blockquote>

A java implementation of the labeler is definitely desirable, similar to the reason why we added the any-sketch java library. Since all EDPs are expected to use the any-sketch an VID labeler library.

The question is whether we should implement the library in Java directly, or call the c++ version via JNI. I would vote for the former, but it is more time consuming.

</blockquote></details>

@wangyaopw - another thing to keep in mind is that Riegeli doesn't have a java reader so we'd have to port that impl as well if we want a non-jni version

___

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/38)* status: 10 of 40 files reviewed, 3 unresolved discussions (waiting on @robinsons, @stevenwarejones, @tcsnfkx, @wangyaopw, and @wliue)

*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/38#-N2sUBNbBkw_TzooTFBn:-N3_U0i9-Gui8ISDyKpB:b-j810nb) (no related file):*

@wangyaopw - another thing to keep in mind is that Riegeli doesn't have a java reader so we'd have to port that impl as well if we want a non-jni version
</blockquote></details>

FYI there is a WIP PR for the Riegeli project adding its own Java wrapper, but there hasn't been progress for a bit: https://github.com/google/riegeli/pull/21

___

wangyaopw (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/38)* status: 10 of 40 files reviewed, 3 unresolved discussions (waiting on @robinsons, @SanjayVas, @stevenwarejones, @tcsnfkx, @wangyaopw, and @wliue)

*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/38#-N2sUBNbBkw_TzooTFBn:-N3_V-A44dCGssBcRyCb:biir9v7) (no related file):*
<details><summary><i>Previously, SanjayVas (Sanjay Vasandani) wrote…</i></summary><blockquote>

FYI there is a WIP PR for the Riegeli project adding its own Java wrapper, but there hasn't been progress for a bit: https://github.com/google/riegeli/pull/21
</blockquote></details>

Even without the Riegeli java reader, we can still just read the model from the textproto file. I assume the job will only read the VID model once in the beginning, so reading from a less efficient file format probably is not a big deal.

___

wliue (commented):
> Reviewed all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/38)* status: 10 of 40 files reviewed, 3 unresolved discussions (waiting on @robinsons, @SanjayVas, @stevenwarejones, @tcsnfkx, and @wangyaopw)

*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/38#-N2sUBNbBkw_TzooTFBn:-N3_YyD01tszYej3UQvH:b8t98fr) (no related file):*
<details><summary><i>Previously, wangyaopw (Yao Wang) wrote…</i></summary><blockquote>

Even without the Riegeli java reader, we can still just read the model from the textproto file. I assume the job will only read the VID model once in the beginning, so reading from a less efficient file format probably is not a big deal.
</blockquote></details>

Riegeli is only used in the initial model loading. And we can add support for other format if needed. 

___

PR #72, opened 2023-06-06
Init new classes

Marco-Premier:


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/72)

PR #76, opened 2023-08-25
Vid selector cpp

Marco-Premier:
C++ version for the VidModelSelector

Note: There are no comments in the code as per the others cpp files we have.
Is that ok?


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/76)

wliue (changes requested):
Which cpp files are you referring to? Cpp files can and should have comments as needed, e.g. 
https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/main/src/main/cc/wfa/virtual_people/core/model/branch_node_impl.h
https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/main/src/main/cc/wfa/virtual_people/core/model/branch_node_impl.cc

> Reviewed 1 of 19 files at r1.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/76#-:-NczyNUf4wJpiekUED2f:b-qeg4is)* status: 1 of 19 files reviewed, 1 unresolved discussion (waiting on @kungfucraig, @Marco-Premier, @SanjayVas, and @stevenwarejones)

___
*[`src/main/cc/wfa/virtual_people/core/selector/testing/testdata/BUILD.bazel` line 1 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/76#-Nd-04lp9Po8uyjgQ7t0:-Nd-04lp9Po8uyjgQ7t1:b-dgtssi) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/9731196c882e6a5666c747505b1834e96aa8abb2/src/main/cc/wfa/virtual_people/core/selector/testing/testdata/BUILD.bazel#L1)):*
> ```Starlark
> package(
> ```

Kotlin and c++ use the same test proto files, correct? Can we move these files to a common dir, similar to 
https://github.com/world-federation-of-advertisers/virtual-people-core-serving/tree/main/src/test/proto/wfa/virtual_people/core/labeler/test_data

Marco-Premier (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/76)* status: 0 of 30 files reviewed, 1 unresolved discussion (waiting on @kungfucraig, @SanjayVas, @stevenwarejones, and @wliue)

___
*[`src/main/cc/wfa/virtual_people/core/selector/testing/testdata/BUILD.bazel` line 1 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/76#-Nd-04lp9Po8uyjgQ7t0:-Nd-nIoqD4pXCYXIc8XF:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/9731196c882e6a5666c747505b1834e96aa8abb2/src/main/cc/wfa/virtual_people/core/selector/testing/testdata/BUILD.bazel#L1)):*
<details><summary><i>Previously, wliue (Wei Liu) wrote…</i></summary><blockquote>

Kotlin and c++ use the same test proto files, correct? Can we move these files to a common dir, similar to 
https://github.com/world-federation-of-advertisers/virtual-people-core-serving/tree/main/src/test/proto/wfa/virtual_people/core/labeler/test_data
</blockquote></details>

Done.

SanjayVas (changes requested):
> Reviewed 1 of 19 files at r1, 2 of 5 files at r2, 1 of 26 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/76)* status: 4 of 30 files reviewed, 11 unresolved discussions (waiting on @kungfucraig, @Marco-Premier, @stevenwarejones, and @wliue)

___
*[`src/main/cc/wfa/virtual_people/core/selector/BUILD.bazel` line 8 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/76#-Nd1JTrp65OGmu02olDf:-Nd1JTrp65OGmu02olDg:b8x644l) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/2ece944093b83c406dc8956c838d33b4f39fc0e6/src/main/cc/wfa/virtual_people/core/selector/BUILD.bazel#L8)):*
> ```Starlark
> _INCLUDE_PREFIX = "/src/main/cc"
> 
> cc_proto_library(
> ```

These should go in `src/main/proto/wfa/measurement/api/v2alpha/BUILD.bazel` in this repo.
___
*[`src/main/cc/wfa/virtual_people/core/selector/lru_cache.h` line 33 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/76#-Nd1LJGd-oltvL2ld5O6:-Nd1LJGd-oltvL2ld5O7:bdxaw00) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/2ece944093b83c406dc8956c838d33b4f39fc0e6/src/main/cc/wfa/virtual_people/core/selector/lru_cache.h#L33)):*
> ```objc
> };
> 
> class LruCache {
> ```

It might be worth making this generic and moving it to common-cpp.
___
*[`src/main/cc/wfa/virtual_people/core/selector/lru_cache.h` line 33 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/76#-Nd1O-5v54uaKj9Q-jfv:-Nd1O-5v54uaKj9Q-jfw:bu92w0s) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/2ece944093b83c406dc8956c838d33b4f39fc0e6/src/main/cc/wfa/virtual_people/core/selector/lru_cache.h#L33)):*
> ```objc
> };
> 
> class LruCache {
> ```

Document this type. I believe the important aspects are that it's a least recently used (LRU) cache with a fixed maximum number of elements.
___
*[`src/main/cc/wfa/virtual_people/core/selector/lru_cache.h` line 35 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/76#-Nd1N88UCiKYCuBzuNRt:-Nd1N88UCiKYCuBzuNRu:bm5cifc) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/2ece944093b83c406dc8956c838d33b4f39fc0e6/src/main/cc/wfa/virtual_people/core/selector/lru_cache.h#L35)):*
> ```objc
> class LruCache {
>  public:
>   explicit LruCache(int n);
> ```

Use meaningful names.



_Suggestion:_
```objc
max_elements
```
___
*[`src/main/cc/wfa/virtual_people/core/selector/lru_cache.h` line 37 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/76#-Nd1NGelAzPuZe3ytItk:-Nd1NGelAzPuZe3ytItl:b-olemi6) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/2ece944093b83c406dc8956c838d33b4f39fc0e6/src/main/cc/wfa/virtual_people/core/selector/lru_cache.h#L37)):*
> ```objc
>   explicit LruCache(int n);
> 
>   void Add(const std::tm& key, const std::vector<ModelReleasePercentile>& data);
> ```

If this is meant to be equivalent to `google.protobuf.Timestamp`, I believe the preferred C++ data structure is `absl::Time`.



_Code quote:_
```objc
std::tm
```
___
*[`src/main/cc/wfa/virtual_people/core/selector/lru_cache.h` line 42 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/76#-Nd1PO0b-yM70eDQIYe8:-Nd1PO0c9oHiQldn1y6Y:b9rmec5) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/2ece944093b83c406dc8956c838d33b4f39fc0e6/src/main/cc/wfa/virtual_people/core/selector/lru_cache.h#L42)):*
> ```objc
> 
>  private:
>   absl::flat_hash_map<std::string, std::vector<ModelReleasePercentile>>
> ```

Why do we need to use a string key? The map implementations should support arbitrary key types. You just might have to provide hashing/comparison functions.



_Code quote:_
```objc
std::string
```
___
*[`src/main/cc/wfa/virtual_people/core/selector/lru_cache.cc` line 28 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/76#-Nd1RwFi-ttx0FmH4W_X:-Nd1RwFi-ttx0FmH4W_Y:b-m5l5gq) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/2ece944093b83c406dc8956c838d33b4f39fc0e6/src/main/cc/wfa/virtual_people/core/selector/lru_cache.cc#L28)):*
> ```cpp
> }
> 
> // Add a new entry into the cache. If the cache is full, the oldest element is
> ```

Function documentation generally goes on the declaration, not the definition. See https://google.github.io/styleguide/cppguide.html#Function_Comments
___
*[`src/main/cc/wfa/virtual_people/core/selector/vid_model_selector.h` line 36 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/76#-Nd1S6NE9x09WZeNK_aJ:-Nd1S6NF0faZ9hMNGbRo:bum8tp3) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/2ece944093b83c406dc8956c838d33b4f39fc0e6/src/main/cc/wfa/virtual_people/core/selector/vid_model_selector.h#L36)):*
> ```objc
>   VidModelSelector(const ModelLine& model_line,
>                    const std::vector<ModelRollout>& model_rollouts);
>   std::optional<std::string> GetModelRelease(const LabelerInput* labeler_input);
> ```

If the parameter cannot be null it should be a const reference instead of a const pointer.



_Code quote:_
```objc
const LabelerInput* labeler_input
```
___
*[`src/main/cc/wfa/virtual_people/core/selector/vid_model_selector.cc` line 26 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/76#-Nd1TI0d4TiMMaN0esB3:-Nd1TI0d4TiMMaN0esB4:b4l4g2p) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/2ece944093b83c406dc8956c838d33b4f39fc0e6/src/main/cc/wfa/virtual_people/core/selector/vid_model_selector.cc#L26)):*
> ```cpp
> namespace wfa_virtual_people {
> 
> const int CACHE_SIZE = 60;
> ```

Follow style guide for naming constants: https://google.github.io/styleguide/cppguide.html#Constant_Names



_Suggestion:_
```cpp
kCacheSize
```
___
*[`src/main/cc/wfa/virtual_people/core/selector/vid_model_selector.cc` line 91 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/76#-Nd1SX2UBNTsl4VSPSqX:-Nd1SX2UBNTsl4VSPSqY:b-z48mwo) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/2ece944093b83c406dc8956c838d33b4f39fc0e6/src/main/cc/wfa/virtual_people/core/selector/vid_model_selector.cc#L91)):*
> ```cpp
> 
> // Class constructor.
> // Throws an exception if model_line name is unspecified or invalid and if
> ```

Our C++ code must be exception-free[^1]. We use return values (see absl::Status) instead of exceptions.

For constructors, this often means making the real constructor private and instead having a factory function so it can return `absl::StatusOr`.

[^1]: https://google.github.io/styleguide/cppguide.html#Exceptions



_Code quote:_
```cpp
Throws an exception
```
___
*[`src/main/cc/wfa/virtual_people/core/selector/testing/testdata/BUILD.bazel` line 1 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/76#-Nd-04lp9Po8uyjgQ7t0:-Nd1UJVM-qfG9ay696gp:b-g045df) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/9731196c882e6a5666c747505b1834e96aa8abb2/src/main/cc/wfa/virtual_people/core/selector/testing/testdata/BUILD.bazel#L1)):*
<details><summary><i>Previously, Marco-Premier (marcopremier) wrote…</i></summary><blockquote>

Done.
</blockquote></details>

Nothing should be in `src/test/proto`, as that would be for tests written in the protobuf definition language (which don't exist). Perhaps something under `src/main/resources` instead.

Marco-Premier (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/76)* status: 5 of 31 files reviewed, 10 unresolved discussions (waiting on @kungfucraig, @SanjayVas, @stevenwarejones, and @wliue)

___
*[`src/main/cc/wfa/virtual_people/core/selector/BUILD.bazel` line 8 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/76#-Nd1JTrp65OGmu02olDf:-NdAUvY48wyuwCFgA_Xi:b-tcw9o0) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/2ece944093b83c406dc8956c838d33b4f39fc0e6/src/main/cc/wfa/virtual_people/core/selector/BUILD.bazel#L8)):*
<details><summary><i>Previously, SanjayVas (Sanjay Vasandani) wrote…</i></summary><blockquote>

Yes. All of the `*proto_library` targets go under the src/main/proto tree. Feel free to handle those in a separate PR.
</blockquote></details>

Right, added a TODO
___
*[`src/main/cc/wfa/virtual_people/core/selector/lru_cache.h` line 37 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/76#-Nd1NGelAzPuZe3ytItk:-NdAvdKW1Of16KMek-Ms:b-jtielj) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/2ece944093b83c406dc8956c838d33b4f39fc0e6/src/main/cc/wfa/virtual_people/core/selector/lru_cache.h#L37)):*
<details><summary><i>Previously, SanjayVas (Sanjay Vasandani) wrote…</i></summary><blockquote>

Ah, so this is mean to be equivalent to the protobuf type `google.type.Date`. I believe the matching structure for that is `absl::CivilDay`. This structure should also be intrinsically supported by absl::Hash so you don't need to define a separate hash function.

The absl time library is to `<ctime>` what the Java 8 `java.time` is to pre-Java 8 time. It should always be used in lieu of anything from `<ctime>`.
</blockquote></details>

Thanks a mil for this :)
___
*[`src/main/cc/wfa/virtual_people/core/selector/vid_model_selector.h` line 41 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/76#-Nd6u3bM-X8us3_IeKli:-NdFlYbGDM9tZG3N_4Kf:b-et34bx) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/54ff072cdfe7a0ac58194f5b52390d327d874dd9/src/main/cc/wfa/virtual_people/core/selector/vid_model_selector.h#L41)):*
<details><summary><i>Previously, SanjayVas (Sanjay Vasandani) wrote…</i></summary><blockquote>

Does VidModelSelector need to be heap-allocated? You shouldn't need to do this just to avoid a copy. See https://abseil.io/tips/166
</blockquote></details>

That didn't work since VidModelSelector constructor is not movable. I did explicit its (movable) declaration now so I could get rid of the unique_ptr
___
*[`src/main/cc/wfa/virtual_people/core/selector/vid_model_selector.h` line 45 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/76#-Nd6wEfs-KKBIq0WoFEs:-NdF0EW23Lqgyhwdby5O:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/54ff072cdfe7a0ac58194f5b52390d327d874dd9/src/main/cc/wfa/virtual_people/core/selector/vid_model_selector.h#L45)):*
<details><summary><i>Previously, SanjayVas (Sanjay Vasandani) wrote…</i></summary><blockquote>

Why is this wrapped in a unique_ptr? You shouldn't need this just to avoid a copy in C++17.
</blockquote></details>

Done.
___
*[`src/main/cc/wfa/virtual_people/core/selector/vid_model_selector.h` line 48 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/76#-Nd6qjfy1XzXt_t9MxzA:-NdFllzoB67K2Q7-yLRx:bu9eyo5) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/54ff072cdfe7a0ac58194f5b52390d327d874dd9/src/main/cc/wfa/virtual_people/core/selector/vid_model_selector.h#L48)):*
<details><summary><i>Previously, SanjayVas (Sanjay Vasandani) wrote…</i></summary><blockquote>

Shouldn't this be private to prevent anyone else from calling it?
</blockquote></details>

yes it was public to allow the ```Build``` method to invoke the ```make_unique``` method.
Now I moved it to private
___
*[`src/main/cc/wfa/virtual_people/core/selector/vid_model_selector.h` line 122 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/76#-Nd6rlTiDNEXfqVyfkRo:-NdBBn6kBEHHn9Ccpa7L:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/54ff072cdfe7a0ac58194f5b52390d327d874dd9/src/main/cc/wfa/virtual_people/core/selector/vid_model_selector.h#L122)):*
<details><summary><i>Previously, SanjayVas (Sanjay Vasandani) wrote…</i></summary><blockquote>

There's no need to define private static methods in header files. Instead, give them internal linkage by declaring them in an unnamed namespace in the `.cc` file. See https://google.github.io/styleguide/cppguide.html#Internal_Linkage
</blockquote></details>

Done.
___
*[`src/main/cc/wfa/virtual_people/core/selector/vid_model_selector.h` line 122 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/76#-Nd6s0YB3lg_NGnFsjmS:-NdEqOn-5vBxpWItl2VS:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/54ff072cdfe7a0ac58194f5b52390d327d874dd9/src/main/cc/wfa/virtual_people/core/selector/vid_model_selector.h#L122)):*
<details><summary><i>Previously, SanjayVas (Sanjay Vasandani) wrote…</i></summary><blockquote>

See https://abseil.io/tips/1. We're already on C++17 so we can use `std::string_view` rather than `absl::string_view`.
</blockquote></details>

Done.
___
*[`src/main/cc/wfa/virtual_people/core/selector/vid_model_selector.cc` line 28 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/76#-Nd6r7wv7j-msXShZ4DT:-NdBC3IrAqZEQue899V2:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/54ff072cdfe7a0ac58194f5b52390d327d874dd9/src/main/cc/wfa/virtual_people/core/selector/vid_model_selector.cc#L28)):*
<details><summary><i>Previously, SanjayVas (Sanjay Vasandani) wrote…</i></summary><blockquote>

Wrap these constants in an unnamed namespace to give them internal linkage. See https://google.github.io/styleguide/cppguide.html#Internal_Linkage
</blockquote></details>

Done.
___
*[`src/main/cc/wfa/virtual_people/core/selector/vid_model_selector.cc` line 30 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/76#-Nd6r52q3RuYYnwYQBMn:-NdBAUuNBHi9FLSo1aFH:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/54ff072cdfe7a0ac58194f5b52390d327d874dd9/src/main/cc/wfa/virtual_people/core/selector/vid_model_selector.cc#L30)):*
<details><summary><i>Previously, SanjayVas (Sanjay Vasandani) wrote…</i></summary><blockquote>

Same note here re:naming constants.
</blockquote></details>

Done.
___
*[`src/main/cc/wfa/virtual_people/core/selector/vid_model_selector.cc` line 120 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/76#-Nd7-obG6AdG5qwKQJOG:-NdEvLcQCuMTfju3pgeV:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/54ff072cdfe7a0ac58194f5b52390d327d874dd9/src/main/cc/wfa/virtual_people/core/selector/vid_model_selector.cc#L120)):*
<details><summary><i>Previously, SanjayVas (Sanjay Vasandani) wrote…</i></summary><blockquote>

Use appropriate structures for date/time operations rather than dealing with primitive types. (This is similar to a comment I had left on #73).
</blockquote></details>

Done.

SanjayVas (changes requested):
> Reviewed 3 of 6 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/76)* status: 4 of 31 files reviewed, 11 unresolved discussions (waiting on @kungfucraig, @Marco-Premier, @stevenwarejones, and @wliue)

___
*[`src/main/cc/wfa/virtual_people/core/selector/lru_cache.h` line 63 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/76#-NdH0Otb3hgxgx3vtu0T:-NdH0Otb3hgxgx3vtu0U:bt3gw0x) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/b421b5d924906f553411fda50b6a2f3f25e35556/src/main/cc/wfa/virtual_people/core/selector/lru_cache.h#L63)):*
> ```objc
>  private:
>   absl::flat_hash_map<absl::CivilDay, std::vector<ModelReleasePercentile>,
>                       CivilDayHash, CivilDayEqual>
> ```

I don't think you actually need to specify the `Eq` and `Hash` template parameters. Since absl::Hash should have inherent support for CivilDay, you should be able to use the defaults.



_Suggestion:_
```objc
  absl::flat_hash_map<absl::CivilDay, std::vector<ModelReleasePercentile>>
```
___
*[`src/main/cc/wfa/virtual_people/core/selector/vid_model_selector.h` line 35 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/76#-NdH8VR48uogz2vnNeHQ:-NdH8VR48uogz2vnNeHR:b-2bwc9c) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/b421b5d924906f553411fda50b6a2f3f25e35556/src/main/cc/wfa/virtual_people/core/selector/vid_model_selector.h#L35)):*
> ```objc
> using ::wfa::measurement::api::v2alpha::ModelRollout;
> 
> class VidModelSelector {
> ```

Make sure to declare any method that does not mutate instance state as `const`.

Note that you should consider whether the cache member is actually instance state. If not, use the `mutable` keyword.
___
*[`src/main/cc/wfa/virtual_people/core/selector/vid_model_selector.h` line 48 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/76#-NdH2d2B4ZWdh8CgQb_a:-NdH2d2B4ZWdh8CgQb_b:b-a7hha9) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/b421b5d924906f553411fda50b6a2f3f25e35556/src/main/cc/wfa/virtual_people/core/selector/vid_model_selector.h#L48)):*
> ```objc
>       const LabelerInput& labeler_input);
> 
>   // Move constructor
> ```

If you declare a move constructor you should also declare a move assignment operator (`VidModelSelector& operator=(VidModelSelector&& rhs)`).

Note that you can specify the `=default` implementation if that works for your class design.

See https://google.github.io/styleguide/cppguide.html#Copyable_Movable_Types
___
*[`src/main/cc/wfa/virtual_people/core/selector/vid_model_selector.h` line 52 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/76#-NdH9vD_6-s0pNZ6za50:-NdH9vD_6-s0pNZ6za51:b-94gwh4) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/b421b5d924906f553411fda50b6a2f3f25e35556/src/main/cc/wfa/virtual_people/core/selector/vid_model_selector.h#L52)):*
> ```objc
> 
>  private:
>   ModelLine model_line;
> ```

Are these intended to be mutable?
___
*[`src/main/cc/wfa/virtual_people/core/selector/vid_model_selector.h` line 52 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/76#-NdHCox9CMQM-0Kuppv5:-NdHCox9CMQM-0Kuppv6:boo3g2z) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/b421b5d924906f553411fda50b6a2f3f25e35556/src/main/cc/wfa/virtual_people/core/selector/vid_model_selector.h#L52)):*
> ```objc
> 
>  private:
>   ModelLine model_line;
> ```

Private data members of classes should be named with a trailing underscore. See https://google.github.io/styleguide/cppguide.html#Variable_Names
___
*[`src/main/cc/wfa/virtual_people/core/selector/vid_model_selector.h` line 88 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/76#-NdHD2zL1oruJmoDzbjN:-NdHD2zL1oruJmoDzbjO:bb2c092) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/b421b5d924906f553411fda50b6a2f3f25e35556/src/main/cc/wfa/virtual_people/core/selector/vid_model_selector.h#L88)):*
> ```objc
>   // Returns the percentage of events that this ModelRollout must label for the
>   // given `event_date_utc`.
>   double CalculatePercentageAdoption(absl::CivilDay& event_date_utc,
> ```

I assume this parameter should be const, as the function isn't mutating it.



_Suggestion:_
```objc
const absl::CivilDay& event_date_utc
```
___
*[`src/main/cc/wfa/virtual_people/core/selector/vid_model_selector.h` line 98 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/76#-NdHDEGUDCJOoxoGxsdL:-NdHDEGUDCJOoxoGxsdM:undefined) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/b421b5d924906f553411fda50b6a2f3f25e35556/src/main/cc/wfa/virtual_people/core/selector/vid_model_selector.h#L98)):*
> ```objc
>   // >= rollout_period_end_date && !rollout.has_rollout_freeze_date()
>   std::vector<ModelRollout> RetrieveActiveRollouts(
>       absl::CivilDay& event_date_utc);
> ```





_Suggestion:_
```objc
const absl::CivilDay& event_date_utc
```
___
*[`src/main/cc/wfa/virtual_people/core/selector/vid_model_selector.h` line 103 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/76#-NdH7xs163iOEC-cB0wp:-NdH7xs163iOEC-cB0wq:borln8c) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/b421b5d924906f553411fda50b6a2f3f25e35556/src/main/cc/wfa/virtual_people/core/selector/vid_model_selector.h#L103)):*
> ```objc
>   // Converts a given absl::Time into a absl::CivilDay object.
>   // absl::CivilDay is used to compare dates in UTC time.
>   absl::CivilDay TimeUsecToCivilDay(absl::Time time);
> ```

These conversion functions don't sound like they'd depend on any other instance members. If they do not, they should not be declared as instance methods.
___
*[`src/main/cc/wfa/virtual_people/core/selector/vid_model_selector.cc` line 28 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/76#-Nd6r7wv7j-msXShZ4DT:-NdH3gnh7nufNnShA2s6:bp5y3mr) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/54ff072cdfe7a0ac58194f5b52390d327d874dd9/src/main/cc/wfa/virtual_people/core/selector/vid_model_selector.cc#L28)):*
<details><summary><i>Previously, Marco-Premier (marcopremier) wrote…</i></summary><blockquote>

Done.
</blockquote></details>

The internal linkage namespace generally goes inside your other namespace. The structure should look something like this:

```cpp
/* includes */

namespace wfa_virtual_people {

/* `using` statements if needed */

namespace {

/* internal definitions */

}  // namespace

/* public definitions */

}  // namespace wfa_virtual_people
```

Marco-Premier (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/76)* status: 4 of 31 files reviewed, 11 unresolved discussions (waiting on @kungfucraig, @SanjayVas, @stevenwarejones, and @wliue)

___
*[`src/main/cc/wfa/virtual_people/core/selector/lru_cache.h` line 63 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/76#-NdH0Otb3hgxgx3vtu0T:-NdU5oW68Ikhi1b9bDBC:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/b421b5d924906f553411fda50b6a2f3f25e35556/src/main/cc/wfa/virtual_people/core/selector/lru_cache.h#L63)):*
<details><summary><i>Previously, SanjayVas (Sanjay Vasandani) wrote…</i></summary><blockquote>

I don't think you actually need to specify the `Eq` and `Hash` template parameters. Since absl::Hash should have inherent support for CivilDay, you should be able to use the defaults.
</blockquote></details>

Done.
___
*[`src/main/cc/wfa/virtual_people/core/selector/vid_model_selector.h` line 35 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/76#-NdH8VR48uogz2vnNeHQ:-NdUG-ja3jHj3Ia2XdC5:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/b421b5d924906f553411fda50b6a2f3f25e35556/src/main/cc/wfa/virtual_people/core/selector/vid_model_selector.h#L35)):*
<details><summary><i>Previously, SanjayVas (Sanjay Vasandani) wrote…</i></summary><blockquote>

Make sure to declare any method that does not mutate instance state as `const`.

Note that you should consider whether the cache member is actually instance state. If not, use the `mutable` keyword.
</blockquote></details>

Done.
___
*[`src/main/cc/wfa/virtual_people/core/selector/vid_model_selector.h` line 48 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/76#-NdH2d2B4ZWdh8CgQb_a:-NdUJ2HfEA7HBaFYovqF:bme0w0d) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/b421b5d924906f553411fda50b6a2f3f25e35556/src/main/cc/wfa/virtual_people/core/selector/vid_model_selector.h#L48)):*
<details><summary><i>Previously, SanjayVas (Sanjay Vasandani) wrote…</i></summary><blockquote>

If you declare a move constructor you should also declare a move assignment operator (`VidModelSelector& operator=(VidModelSelector&& rhs)`).

Note that you can specify the `=default` implementation if that works for your class design.

See https://google.github.io/styleguide/cppguide.html#Copyable_Movable_Types
</blockquote></details>

I don't think I can declare the move assignment operator since ```model_line_``` and ```model_rollouts_``` are marked as ```const```. That prevents me from assigning their values eg.

```
VidModelSelector& VidModelSelector::operator=(VidModelSelector&& other) noexcept {
  if (this != &other) {  
    model_line_ = std::move(other.model_line_); // compiler error
```

So I define the assignment operator as ```deleted```.
Are you happy with this approach?
___
*[`src/main/cc/wfa/virtual_people/core/selector/vid_model_selector.h` line 52 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/76#-NdH9vD_6-s0pNZ6za50:-NdU8d-K6yjEQV5XuTtb:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/b421b5d924906f553411fda50b6a2f3f25e35556/src/main/cc/wfa/virtual_people/core/selector/vid_model_selector.h#L52)):*
<details><summary><i>Previously, SanjayVas (Sanjay Vasandani) wrote…</i></summary><blockquote>

Are these intended to be mutable?
</blockquote></details>

Done.
___
*[`src/main/cc/wfa/virtual_people/core/selector/vid_model_selector.h` line 52 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/76#-NdHCox9CMQM-0Kuppv5:-NdUATWqAAhu9ttIb8rm:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/b421b5d924906f553411fda50b6a2f3f25e35556/src/main/cc/wfa/virtual_people/core/selector/vid_model_selector.h#L52)):*
<details><summary><i>Previously, SanjayVas (Sanjay Vasandani) wrote…</i></summary><blockquote>

Private data members of classes should be named with a trailing underscore. See https://google.github.io/styleguide/cppguide.html#Variable_Names
</blockquote></details>

Done.
___
*[`src/main/cc/wfa/virtual_people/core/selector/vid_model_selector.h` line 88 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/76#-NdHD2zL1oruJmoDzbjN:-NdUBFhy3wj_L_Ef5-5G:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/b421b5d924906f553411fda50b6a2f3f25e35556/src/main/cc/wfa/virtual_people/core/selector/vid_model_selector.h#L88)):*
<details><summary><i>Previously, SanjayVas (Sanjay Vasandani) wrote…</i></summary><blockquote>

I assume this parameter should be const, as the function isn't mutating it.
</blockquote></details>

Done.
___
*[`src/main/cc/wfa/virtual_people/core/selector/vid_model_selector.h` line 98 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/76#-NdHDEGUDCJOoxoGxsdL:-NdUBGUs-Pa74GtLL9jT:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/b421b5d924906f553411fda50b6a2f3f25e35556/src/main/cc/wfa/virtual_people/core/selector/vid_model_selector.h#L98)):*
> ```objc
>   // >= rollout_period_end_date && !rollout.has_rollout_freeze_date()
>   std::vector<ModelRollout> RetrieveActiveRollouts(
>       absl::CivilDay& event_date_utc);
> ```

Done.
___
*[`src/main/cc/wfa/virtual_people/core/selector/vid_model_selector.h` line 103 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/76#-NdH7xs163iOEC-cB0wp:-NdUDM7v9Zv7ZfopAIat:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/b421b5d924906f553411fda50b6a2f3f25e35556/src/main/cc/wfa/virtual_people/core/selector/vid_model_selector.h#L103)):*
<details><summary><i>Previously, SanjayVas (Sanjay Vasandani) wrote…</i></summary><blockquote>

These conversion functions don't sound like they'd depend on any other instance members. If they do not, they should not be declared as instance methods.
</blockquote></details>

Done.
___
*[`src/main/cc/wfa/virtual_people/core/selector/vid_model_selector.cc` line 28 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/76#-Nd6r7wv7j-msXShZ4DT:-NdU6GxnESLhgWP5ciph:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/54ff072cdfe7a0ac58194f5b52390d327d874dd9/src/main/cc/wfa/virtual_people/core/selector/vid_model_selector.cc#L28)):*
<details><summary><i>Previously, SanjayVas (Sanjay Vasandani) wrote…</i></summary><blockquote>

The internal linkage namespace generally goes inside your other namespace. The structure should look something like this:

```cpp
/* includes */

namespace wfa_virtual_people {

/* `using` statements if needed */

namespace {

/* internal definitions */

}  // namespace

/* public definitions */

}  // namespace wfa_virtual_people
```
</blockquote></details>

Done.

SanjayVas (changes requested):
> Reviewed 1 of 19 files at r1, 10 of 26 files at r3, 13 of 20 files at r4, 2 of 6 files at r5, 4 of 4 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/76)* status: all files reviewed, 6 unresolved discussions (waiting on @kungfucraig, @Marco-Premier, @stevenwarejones, and @wliue)

___
*[`src/main/cc/wfa/virtual_people/core/selector/vid_model_selector.h` line 48 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/76#-NdH2d2B4ZWdh8CgQb_a:-NdbHD3l4t12sSHp0GGN:b-eks35t) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/b421b5d924906f553411fda50b6a2f3f25e35556/src/main/cc/wfa/virtual_people/core/selector/vid_model_selector.h#L48)):*
<details><summary><i>Previously, Marco-Premier (marcopremier) wrote…</i></summary><blockquote>

I don't think I can declare the move assignment operator since ```model_line_``` and ```model_rollouts_``` are marked as ```const```. That prevents me from assigning their values eg.

```
VidModelSelector& VidModelSelector::operator=(VidModelSelector&& other) noexcept {
  if (this != &other) {  
    model_line_ = std::move(other.model_line_); // compiler error
```

So I define the assignment operator as ```deleted```.
Are you happy with this approach?
</blockquote></details>

Yes. The comment about having a move assignment operator was for when the type was actually assignable. Now that the data members are`const`, this type is no longer assignable.

See https://abseil.io/tips/177 on type design w.r.t. assignability and mutability.
___
*[`src/main/cc/wfa/virtual_people/core/selector/vid_model_selector.cc` line 185 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/76#-Nd70uK73Kqi1x09Afph:-NdbKJ8h75CbkIxz0beA:b6d1i1) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/54ff072cdfe7a0ac58194f5b52390d327d874dd9/src/main/cc/wfa/virtual_people/core/selector/vid_model_selector.cc#L185)):*
<details><summary><i>Previously, SanjayVas (Sanjay Vasandani) wrote…</i></summary><blockquote>

nit: I think using `emplace_back` here will guarantee there's no extra copy/move. See https://abseil.io/tips/65
</blockquote></details>

I think you slightly misunderstood the usage. `emplace_back` works by doing the allocation/instantiation as part of adding to the collection. To get the benefit, you have to omit the separate instantiation.

```cpp
result.emplace_back(percentage, active_rollout.model_release());
```

If you really want the separate instantiation then use `push_back` as you had before.

See https://abseil.io/tips/112, which I now realize may actually prefer `push_back` for readability unless there's a significant performance gain.
___
*[`src/main/cc/wfa/virtual_people/core/selector/vid_model_selector.cc` line 39 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/76#-NdbGsG0BOHd3cGICQUC:-NdbGsG0BOHd3cGICQUD:b-1o0dek) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/6f6916c8b5bbdeb0b01f76522385ae625ee86e58/src/main/cc/wfa/virtual_people/core/selector/vid_model_selector.cc#L39)):*
> ```cpp
> // Returns the model_line_id from the given resource name.
> // Returns an empty string if not model_line_id is found.
> static std::string_view ReadModelLine(std::string_view input) {
> ```

You shouldn't need the `static` keyword on functions that don't belong to a type.



_Code quote:_
```cpp
static 
```
___
*[`src/test/cc/wfa/virtual_people/core/selector/BUILD.bazel` line 13 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/76#-NdbNcZaAhCXhe4f2mbP:-NdbNcZaAhCXhe4f2mbQ:b-ogq6z8) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/6f6916c8b5bbdeb0b01f76522385ae625ee86e58/src/test/cc/wfa/virtual_people/core/selector/BUILD.bazel#L13)):*
> ```Starlark
>     name = "model_rollout_cc_proto",
>     deps = ["@wfa_measurement_proto//src/main/proto/wfa/measurement/api/v2alpha:model_rollout_proto"],
> )
> ```

I thought these were already moved



_Code quote:_
```Starlark
cc_proto_library(
    name = "model_line_cc_proto",
    deps = ["@wfa_measurement_proto//src/main/proto/wfa/measurement/api/v2alpha:model_line_proto"],
)

cc_proto_library(
    name = "model_rollout_cc_proto",
    deps = ["@wfa_measurement_proto//src/main/proto/wfa/measurement/api/v2alpha:model_rollout_proto"],
)
```
___
*[`src/main/cc/wfa/virtual_people/core/selector/testing/testdata/BUILD.bazel` line 1 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/76#-Nd-04lp9Po8uyjgQ7t0:-NdbNMQq-Sv83y8XXPSR:b-8x5427) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/9731196c882e6a5666c747505b1834e96aa8abb2/src/main/cc/wfa/virtual_people/core/selector/testing/testdata/BUILD.bazel#L1)):*
<details><summary><i>Previously, Marco-Premier (marcopremier) wrote…</i></summary><blockquote>

Right. Shall I move also the other existing proto files for the same reason?

https://github.com/world-federation-of-advertisers/virtual-people-core-serving/tree/main/src/test/proto/wfa/virtual_people/core/labeler/test_data
</blockquote></details>

Yes. thanks.

Marco-Premier (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/76)* status: 29 of 77 files reviewed, 4 unresolved discussions (waiting on @kungfucraig, @SanjayVas, @stevenwarejones, and @wliue)

___
*[`src/main/cc/wfa/virtual_people/core/selector/vid_model_selector.cc` line 185 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/76#-Nd70uK73Kqi1x09Afph:-NdfCEcH8NqhMc4MDEbb:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/54ff072cdfe7a0ac58194f5b52390d327d874dd9/src/main/cc/wfa/virtual_people/core/selector/vid_model_selector.cc#L185)):*
<details><summary><i>Previously, SanjayVas (Sanjay Vasandani) wrote…</i></summary><blockquote>

I think you slightly misunderstood the usage. `emplace_back` works by doing the allocation/instantiation as part of adding to the collection. To get the benefit, you have to omit the separate instantiation.

```cpp
result.emplace_back(percentage, active_rollout.model_release());
```

If you really want the separate instantiation then use `push_back` as you had before.

See https://abseil.io/tips/112, which I now realize may actually prefer `push_back` for readability unless there's a significant performance gain.
</blockquote></details>

Done.
___
*[`src/main/cc/wfa/virtual_people/core/selector/vid_model_selector.cc` line 39 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/76#-NdbGsG0BOHd3cGICQUC:-NdfAop76SgKdqTC_OYZ:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/6f6916c8b5bbdeb0b01f76522385ae625ee86e58/src/main/cc/wfa/virtual_people/core/selector/vid_model_selector.cc#L39)):*
<details><summary><i>Previously, SanjayVas (Sanjay Vasandani) wrote…</i></summary><blockquote>

You shouldn't need the `static` keyword on functions that don't belong to a type.
</blockquote></details>

Done.
___
*[`src/test/cc/wfa/virtual_people/core/selector/BUILD.bazel` line 13 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/76#-NdbNcZaAhCXhe4f2mbP:-NdfCmffBpyzDneOlW6e:b-m2g57d) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/6f6916c8b5bbdeb0b01f76522385ae625ee86e58/src/test/cc/wfa/virtual_people/core/selector/BUILD.bazel#L13)):*
<details><summary><i>Previously, SanjayVas (Sanjay Vasandani) wrote…</i></summary><blockquote>

I thought these were already moved
</blockquote></details>

Yes, my bad
___
*[`src/main/cc/wfa/virtual_people/core/selector/testing/testdata/BUILD.bazel` line 1 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/76#-Nd-04lp9Po8uyjgQ7t0:-NdfEVu328fb3RDiKY-G:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/9731196c882e6a5666c747505b1834e96aa8abb2/src/main/cc/wfa/virtual_people/core/selector/testing/testdata/BUILD.bazel#L1)):*
<details><summary><i>Previously, SanjayVas (Sanjay Vasandani) wrote…</i></summary><blockquote>

Yes. thanks.
</blockquote></details>

Done.

SanjayVas (changes requested):
> Reviewed 48 of 48 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/76)* status: all files reviewed, 3 unresolved discussions (waiting on @kungfucraig, @Marco-Premier, @stevenwarejones, and @wliue)

___
*[`src/main/cc/wfa/virtual_people/core/selector/vid_model_selector.cc` line 185 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/76#-Nd70uK73Kqi1x09Afph:-Ndfb9tr4jm7idfkcvcW:byqsbcv) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/54ff072cdfe7a0ac58194f5b52390d327d874dd9/src/main/cc/wfa/virtual_people/core/selector/vid_model_selector.cc#L185)):*
<details><summary><i>Previously, Marco-Premier (marcopremier) wrote…</i></summary><blockquote>

Done.
</blockquote></details>

You should be passing the args to the `ModelReleasePercentile` constructor to `emplace_back`, not an already-constructed `ModelReleasePercentile` instance. Did the syntax I specified not compile for some reason? 
___
*[`src/main/cc/wfa/virtual_people/core/selector/vid_model_selector.cc` line 39 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/76#-NdbGsG0BOHd3cGICQUC:-NdfawtrDxIIiBFCb3Nb:b-7gcmct) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/6f6916c8b5bbdeb0b01f76522385ae625ee86e58/src/main/cc/wfa/virtual_people/core/selector/vid_model_selector.cc#L39)):*
<details><summary><i>Previously, Marco-Premier (marcopremier) wrote…</i></summary><blockquote>

Done.
</blockquote></details>

This also applies to the other functions in this namespace.

Marco-Premier (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/76)* status: 75 of 77 files reviewed, 3 unresolved discussions (waiting on @kungfucraig, @SanjayVas, @stevenwarejones, and @wliue)

___
*[`src/main/cc/wfa/virtual_people/core/selector/vid_model_selector.cc` line 185 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/76#-Nd70uK73Kqi1x09Afph:-NdioFkk0QYrjmy0h0Fj:brvcbm) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/54ff072cdfe7a0ac58194f5b52390d327d874dd9/src/main/cc/wfa/virtual_people/core/selector/vid_model_selector.cc#L185)):*
<details><summary><i>Previously, SanjayVas (Sanjay Vasandani) wrote…</i></summary><blockquote>

You should be passing the args to the `ModelReleasePercentile` constructor to `emplace_back`, not an already-constructed `ModelReleasePercentile` instance. Did the syntax I specified not compile for some reason?
</blockquote></details>

Turns out that c++ doesn't generate a constructor for this struct and that's why ```emplace_back``` wasn't working (yes compiler was raising an exception: ```no matching constructor for initialization of 'wfa_virtual_people::ModelReleasePercentile'```

I added he constructor to ```ModelReleasePercentile``` and now it works
___
*[`src/main/cc/wfa/virtual_people/core/selector/vid_model_selector.cc` line 39 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/76#-NdbGsG0BOHd3cGICQUC:-NdilCfD1k-8ioocHTDj:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/6f6916c8b5bbdeb0b01f76522385ae625ee86e58/src/main/cc/wfa/virtual_people/core/selector/vid_model_selector.cc#L39)):*
<details><summary><i>Previously, SanjayVas (Sanjay Vasandani) wrote…</i></summary><blockquote>

This also applies to the other functions in this namespace.
</blockquote></details>

Done.

SanjayVas (approved):
> Reviewed 2 of 2 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/76)* status: all files reviewed, 1 unresolved discussion (waiting on @kungfucraig, @stevenwarejones, and @wliue)

stevenwarejones (approved):
> Reviewed 1 of 19 files at r1, 10 of 26 files at r3, 13 of 20 files at r4, 2 of 6 files at r5, 2 of 4 files at r6, 47 of 48 files at r7, 2 of 2 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/76)* status: all files reviewed, 1 unresolved discussion (waiting on @kungfucraig and @wliue)

wliue (changes requested):
> Reviewed 1 of 19 files at r1, 10 of 26 files at r3, 13 of 20 files at r4, 2 of 6 files at r5, 1 of 4 files at r6, 47 of 48 files at r7, 2 of 2 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/76)* status: all files reviewed, 3 unresolved discussions (waiting on @Marco-Premier)

___
*[`src/main/cc/wfa/virtual_people/core/selector/lru_cache.h` line 38 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/76#-Ne2QoI2-5LnHoCa_Iq1:-Ne2QoI38-zSW1YBOC34:bg9t88e) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/16f3bb73da97e0e71d63d2efe4772a46483960e5/src/main/cc/wfa/virtual_people/core/selector/lru_cache.h#L38)):*
> ```objc
> };
> 
> // TODO(@marcopremier): Move this class in common-cpp
> ```

We need to add proper test coverage when move this to common-cpp.



_Code quote:_
```objc
// TODO(@marcopremier): Move this class in common-cpp
```
___
*[`src/main/cc/wfa/virtual_people/core/selector/lru_cache.cc` line 31 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/76#-Ne2N_ng3MbYgDq359ax:-Ne2N_nh90F6QpKRJOTf:b-3dtdli) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/16f3bb73da97e0e71d63d2efe4772a46483960e5/src/main/cc/wfa/virtual_people/core/selector/lru_cache.cc#L31)):*
> ```cpp
>     }
>     cache_data.erase(*oldest);
>     access_order.erase(oldest);
> ```

Looks like this removes the oldest key by value. I think normally LRU removes the least recently used entry, i.e. the one at the end of access_order. 



_Code quote:_
```cpp
    auto oldest = access_order.begin();
    for (auto idx = access_order.begin(); idx != access_order.end(); ++idx) {
      if (*idx < *oldest) {
        oldest = idx;
      }
    }
    cache_data.erase(*oldest);
    access_order.erase(oldest);
```

Marco-Premier (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/76)* status: 76 of 77 files reviewed, 3 unresolved discussions (waiting on @Marco-Premier, @SanjayVas, @stevenwarejones, and @wliue)

___
*[`src/main/cc/wfa/virtual_people/core/selector/lru_cache.h` line 38 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/76#-Ne2QoI2-5LnHoCa_Iq1:-Ne3MsNVAGULy74WAUY-:bhy9os6) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/16f3bb73da97e0e71d63d2efe4772a46483960e5/src/main/cc/wfa/virtual_people/core/selector/lru_cache.h#L38)):*
<details><summary><i>Previously, wliue (Wei Liu) wrote…</i></summary><blockquote>

We need to add proper test coverage when move this to common-cpp.
</blockquote></details>

Will do
___
*[`src/main/cc/wfa/virtual_people/core/selector/lru_cache.cc` line 31 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/76#-Ne2N_ng3MbYgDq359ax:-Ne3c1frAsW_F99E3zLw:b-aqbwg3) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/16f3bb73da97e0e71d63d2efe4772a46483960e5/src/main/cc/wfa/virtual_people/core/selector/lru_cache.cc#L31)):*
<details><summary><i>Previously, wliue (Wei Liu) wrote…</i></summary><blockquote>

Looks like this removes the oldest key by value. I think normally LRU removes the least recently used entry, i.e. the one at the end of access_order.
</blockquote></details>

you are right @wliue :) thanks. fixed it

wliue (approved):
> Reviewed 1 of 1 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/76)* status: all files reviewed, 1 unresolved discussion (waiting on @Marco-Premier)

Marco-Premier (commented):
> Reviewed 1 of 19 files at r1, 10 of 26 files at r3, 11 of 20 files at r4, 2 of 6 files at r5, 1 of 4 files at r6, 46 of 48 files at r7, 2 of 2 files at r8, 1 of 1 files at r9, 3 of 3 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/76)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Marco-Premier)

PR #84, opened 2026-04-28
Add memoized VID assignment (Kotlin)

alxlyj:
## Summary

Kotlin implementation of memoized VID assignment: Feistel cipher and RankedPopulationNodeImpl with ModelNode wiring. Pass-1 mode is deferred to a follow-up.

### Changes

**Feistel cipher** (`Feistel.kt`):
- Format-preserving bijection [0, N) → [0, N) with zero collisions
- 4-round Feistel network, FarmHash64 round function, recursive cycle-walking (matches C++)

**RankedPopulationNodeImpl** (`RankedPopulationNodeImpl.kt`):
- Ranked path: Feistel bijection for collision-free VID assignment
- Unranked path: `jumpConsistentHash` with exhaustive `when` for DISJOINT / FULL_POOL
- val-only Apply (no var reassignment)
- Single pool enforced via `single()`

**Shared utilities** (`PopulationNodeHelper.kt`):
- `computeVidSeed` — shared seed computation between PopulationNodeImpl and RankedPopulationNodeImpl
- `collapseQuantumLabel` — extracted from PopulationNodeImpl for reuse

**ModelNode wiring** (`ModelNode.kt`):
- Added `RANKED_POPULATION_NODE` case to `build()` dispatch

## Test Plan

- [x] `bazel test //src/test/kotlin/org/wfanet/virtualpeople/core/model/utils:FeistelTest` — **PASSED**
  - Bijectivity (100, 1000, prime 997), determinism, range, edge cases
  - Golden vectors pinned for cross-language parity detection
- [x] `bazel test //src/test/kotlin/org/wfanet/virtualpeople/core/labeler:RankedLabelingIntegrationTest` — **PASSED**
  - Collision-free ranked VIDs (200 events → 200 unique)
  - Unranked DISJOINT and FULL_POOL range validation
  - Rank overflow (local_rank >= ranked_size) falls back to unranked
  - ranked_size=0 hash-based fallback


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/84)

stevenwarejones (changes requested):
The "Changes" section lists Labeler.kt modifications (LabelingMode enum, overloaded label(), pool_assignments copying) that are not in the diff. If those are planned for a follow-up PR, update the description to clarify the scope of  
  this PR vs. what's coming next — otherwise it reads like the changes were intended but missing.                                                                                                                                           
  
  nit: Add an Issue trailer per dev-standards ("Every PR should be associated with an existing Issue").      

  The doc comment states "Must produce identical output to the C++ FeistelPermute for cross-language determinism" but there are no golden test vectors validated against the C++ implementation. Add a test with hardcoded input/output
  pairs that are independently verified against C++ FeistelPermute — this is critical since both labelers must produce identical VIDs for the same inputs. 

> @stevenwarejones reviewed 1 file and made 6 comments.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/84#-:-Os71GOKCOzC-D9lDiJl:b2q223)* status: 1 of 9 files reviewed, 5 unresolved discussions (waiting on alxlyj and Marco-Premier).

___
*[`src/main/kotlin/org/wfanet/virtualpeople/core/model/RankedPopulationNodeImpl.kt` line 68 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/84#-Os7-ily38h5kJ6AIAga:-Os7-ily38h5kJ6AIAgb:b-831v3i) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/e6745b5c8644a3b37262129fe1b580029719aaf1/src/main/kotlin/org/wfanet/virtualpeople/core/model/RankedPopulationNodeImpl.kt#L68)):*
> ```Kotlin
>     if (hasRank && localRank < rankedSize) {
>       // RANKED path: Feistel bijection — zero collisions.
>       virtualPersonId = poolOffset + Feistel.permute(localRank, rankedSize, randomSeed)
> ```

i'd prefer something iterative and avoid recursion. for this, is it okay to check just only until (current >= domainSize)   Cycle-walking is recursive here. For some domain sizes near the half-size boundary this could recurse deeply. Consider making it iterative:    
```                   
    var current = value                                                                                                                                                                                                                     
                                                                                                                                                                                                                                            
    do {                                                                                                                                                                                                                                    
      // apply 4-round Feistel to current                                                                                                                                                                                                   
                                                                                                                                                                                                                                            
      ...                                                                                                                                                                                                                                   
      current = left * half + right                                                                                                                                                                                                         
                                                                                                                                                                                                                                            
    } while (current >= domainSize)                                                                                                                                                                                                         
    return current                                                                                                                                                                                                                                             
```

___
*[`src/main/kotlin/org/wfanet/virtualpeople/core/model/RankedPopulationNodeImpl.kt` line 78 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/84#-Os7-L-T-_BpuXb_cmR6:-Os7-L-T-_BpuXb_cmR7:b-9bg4p5) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/e6745b5c8644a3b37262129fe1b580029719aaf1/src/main/kotlin/org/wfanet/virtualpeople/core/model/RankedPopulationNodeImpl.kt#L78)):*
> ```Kotlin
> 
>       virtualPersonId =
>         if (unrankedMode == UnrankedMode.DISJOINT) {
> ```

use an exhaustive when
___
*[`src/main/kotlin/org/wfanet/virtualpeople/core/model/RankedPopulationNodeImpl.kt` line 86 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/84#-Os73PqNEZGX2mrDpvFp:-Os73PqNEZGX2mrDpvFq:bvh2imp) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/e6745b5c8644a3b37262129fe1b580029719aaf1/src/main/kotlin/org/wfanet/virtualpeople/core/model/RankedPopulationNodeImpl.kt#L86)):*
> ```Kotlin
>         } else {
>           // FULL_POOL: hash into the entire pool.
>           poolOffset + (seed % poolSize)
> ```

shouldn't this use jumpConsistentHash? https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/8a83a76374a958c8e4cdc8976d3dcfab6fbde517/src/main/kotlin/org/wfanet/virtualpeople/core/model/utils/ConsistentHash.kt#L27
___
*[`src/main/kotlin/org/wfanet/virtualpeople/core/model/RankedPopulationNodeImpl.kt` line 108 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/84#-Os728Fo9jhQBFKfUwBJ:-Os728Fo9jhQBFKfUwBK:b-67250y) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/e6745b5c8644a3b37262129fe1b580029719aaf1/src/main/kotlin/org/wfanet/virtualpeople/core/model/RankedPopulationNodeImpl.kt#L108)):*
> ```Kotlin
> 
>   companion object {
>     private fun collapseQuantumLabel(
> ```

This is copied verbatim from PopulationNodeImpl.collapseQuantumLabel. Extract to a shared utility
___
*[`src/main/kotlin/org/wfanet/virtualpeople/core/model/RankedPopulationNodeImpl.kt` line 134 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/84#-Os72JZ0-4P3fCKoK1Fa:-Os72JZ0-4P3fCKoK1Fb:b26qljw) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/e6745b5c8644a3b37262129fe1b580029719aaf1/src/main/kotlin/org/wfanet/virtualpeople/core/model/RankedPopulationNodeImpl.kt#L134)):*
> ```Kotlin
>       require(config.poolsCount > 0) { "RankedPopulationNode must have at least one pool." }
> 
>       val poolOffset = config.poolsList[0].populationOffset.toULong()
> ```

is only only pool allowed here? if so, it should be single() no [0].


<!-- Sent from Reviewable.io -->

stevenwarejones (changes requested):
> @stevenwarejones made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/84)* status: 1 of 9 files reviewed, 7 unresolved discussions (waiting on alxlyj and Marco-Premier).

___
*[`src/main/kotlin/org/wfanet/virtualpeople/core/model/RankedPopulationNodeImpl.kt` line 34 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/84#-Os74dLk0Xqn4q8I_-3Q:-Os74dLlEPUC3qxa40vj:bkhbo9c) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/e6745b5c8644a3b37262129fe1b580029719aaf1/src/main/kotlin/org/wfanet/virtualpeople/core/model/RankedPopulationNodeImpl.kt#L34)):*
> ```Kotlin
>  * Splits VID assignment into ranked (Feistel, collision-free) and unranked (hash-based) sub-ranges.
>  */
> internal class RankedPopulationNodeImpl
> ```

  The integration design doc explicitly describes overflow behavior: "If RankValue is out of range (RankValue >= ranked_size), the labeler treats the event as unranked and assigns a VID using the existing hash-based path (per the node's
   unranked_mode)." This is the safety valve for when ranked capacity is saturated — the system should degrade gracefully to hash-based assignment rather than producing out-of-range Feistel outputs. There's no test covering this path.
  Add a test where local_rank is >= ranked_size (e.g., ranked_size=200, local_rank=250) and verify the VID lands in the unranked range. This is important to lock down because if it breaks silently, overflow accounts would get incorrect 
  VIDs. 

Also, add unit tests for any public function here.
___
*[`src/main/kotlin/org/wfanet/virtualpeople/core/model/RankedPopulationNodeImpl.kt` line 51 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/84#-Os74lI-5GffAjNy8oHF:-Os74lI-5GffAjNy8oHG:b-ewzwj9) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/e6745b5c8644a3b37262129fe1b580029719aaf1/src/main/kotlin/org/wfanet/virtualpeople/core/model/RankedPopulationNodeImpl.kt#L51)):*
> ```Kotlin
>     val activity = VirtualPersonActivity.newBuilder()
> 
>     // Look up pre-computed rank from LabelerInput.rank_assignments.
> ```

  hasRank, localRank, and virtualPersonId are all var but are never reassigned after their initial conditional assignment. Per code style (B1): "If a variable is never reassigned, it must be val. No exceptions." Restructure as:
```                                                                                                                                                                                                                                            
  val rankAssignment = if (event.hasLabelerInput()) {             
    event.labelerInput.rankAssignmentsList.firstOrNull {                                                                                                                                                                                    
      it.poolOffset.toULong() == poolOffset                                                                                                                                                                                                 
    }                                                                                                                                                                                                                                       
  } else null                                                                                                                                                                                                                               
                                                                                                                                                                                                                                            
  val virtualPersonId = if (rankAssignment != null && rankAssignment.localRank.toULong() < rankedSize) {                                                                                                                                    
    poolOffset + Feistel.permute(rankAssignment.localRank.toULong(), rankedSize, randomSeed)                                                                                                                                                
  } else {                                                                                                                                                                                                                                  
    val seed = Hashing.farmHashFingerprint64()                    
      .hashString("$randomSeed${event.actingFingerprint.toULong()}", StandardCharsets.UTF_8)                                                                                                                                                
      .asLong()                                                                                                                                                                                                                             
      .toULong()                                                                                                                                                                                                                            
    if (unrankedMode == UnrankedMode.DISJOINT) {                                                                                                                                                                                            
      val unrankedSize = poolSize - rankedSize                                                                                                                                                                                              
      check(unrankedSize > 0uL) { "DISJOINT mode with ranked_size == pool_size leaves no unranked space." }                                                                                                                                 
      poolOffset + rankedSize + (seed % unrankedSize)                                                                                                                                                                                       
    } else {                                                                                                                                                                                                                                
      poolOffset + (seed % poolSize)                                                                                                                                                                                                        
    }                                                                                                                                                                                                                                       
  }                                                               
       
```



<!-- Sent from Reviewable.io -->

stevenwarejones (changes requested):
> @stevenwarejones made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/84)* status: 1 of 9 files reviewed, 8 unresolved discussions (waiting on alxlyj and Marco-Premier).

___
*[`src/main/kotlin/org/wfanet/virtualpeople/core/model/RankedPopulationNodeImpl.kt` line 71 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/84#-Os75B4r-eQcGuL2MrAp:-Os75B4r-eQcGuL2MrAq:b-b3kfwg) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/e6745b5c8644a3b37262129fe1b580029719aaf1/src/main/kotlin/org/wfanet/virtualpeople/core/model/RankedPopulationNodeImpl.kt#L71)):*
> ```Kotlin
>     } else {
>       // UNRANKED path: hash-based (mode-dependent scope).
>       val seed =
> ```

  The seed computation here (FarmHash64("$randomSeed${actingFingerprint.toULong()}")) is duplicated from PopulationNodeImpl. Both nodes need to produce the same seed for the same event — if either implementation drifts, the unranked    
  path silently produces different VIDs. Extract the seed computation to a shared utility (same place as collapseQuantumLabel — these two nodes share enough logic that a common base or utility object makes sense). Then both
  PopulationNodeImpl and RankedPopulationNodeImpl call the shared function, keeping them in sync by construction rather than by coincidence.

stevenwarejones (changes requested):
> @stevenwarejones made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/84)* status: 1 of 9 files reviewed, 8 unresolved discussions (waiting on alxlyj and Marco-Premier).

___
*[`src/main/kotlin/org/wfanet/virtualpeople/core/model/RankedPopulationNodeImpl.kt` line 71 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/84#-Os75B4r-eQcGuL2MrAp:-Os75Eul0IPYOF2XEE0g:b24gsce) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/e6745b5c8644a3b37262129fe1b580029719aaf1/src/main/kotlin/org/wfanet/virtualpeople/core/model/RankedPopulationNodeImpl.kt#L71)):*

The seed computation here (FarmHash64("$randomSeed${actingFingerprint.toULong()}")) is duplicated from PopulationNodeImpl. Both nodes need to produce the same seed for the same event — if either implementation drifts, the unranked    
  path silently produces different VIDs. Extract the seed computation to a shared utility (same place as collapseQuantumLabel — these two nodes share enough logic that a common base or utility object makes sense). Then both
  PopulationNodeImpl and RankedPopulationNodeImpl call the shared function, keeping them in sync by construction rather than by coincidence.
</blockquote></details>

i commented in a few places on this but as much as possible share functions across both approaches.

stevenwarejones (changes requested):
> @stevenwarejones reviewed 4 files and all commit messages, made 6 comments, and resolved 6 discussions.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/84)* status: 5 of 11 files reviewed, 6 unresolved discussions (waiting on alxlyj and Marco-Premier).

___
*[`src/main/kotlin/org/wfanet/virtualpeople/core/model/RankedPopulationNodeImpl.kt` line 68 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/84#-Os7-ily38h5kJ6AIAga:-Os8To-84MA_yY25S2jL:b-wez195) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/e6745b5c8644a3b37262129fe1b580029719aaf1/src/main/kotlin/org/wfanet/virtualpeople/core/model/RankedPopulationNodeImpl.kt#L68)):*

i'd prefer something iterative and avoid recursion. for this, is it okay to check just only until (current >= domainSize)   Cycle-walking is recursive here. For some domain sizes near the half-size boundary this could recurse deeply. Consider making it iterative:    
```                   
    var current = value                                                                                                                                                                                                                     
                                                                                                                                                                                                                                            
    do {                                                                                                                                                                                                                                    
      // apply 4-round Feistel to current                                                                                                                                                                                                   
                                                                                                                                                                                                                                            
      ...                                                                                                                                                                                                                                   
      current = left * half + right                                                                                                                                                                                                         
                                                                                                                                                                                                                                            
    } while (current >= domainSize)                                                                                                                                                                                                         
    return current
```

</blockquote></details>

this doesn't look taken care of.
___
*[`src/main/kotlin/org/wfanet/virtualpeople/core/model/RankedPopulationNodeImpl.kt` line 134 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/84#-Os72JZ0-4P3fCKoK1Fa:-Os8U1XH8ahaIXXp4FuY:bc3ovko) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/e6745b5c8644a3b37262129fe1b580029719aaf1/src/main/kotlin/org/wfanet/virtualpeople/core/model/RankedPopulationNodeImpl.kt#L134)):*

is only only pool allowed here? if so, it should be single() no [0].
</blockquote></details>

nit: attach a message
```
require(config.poolsCount == 1) { "RankedPopulationNode requires exactly one pool, got ${config.poolsCount}." }                                                                                                                           
  val pool = config.poolsList.single()  
```

___
*[`src/main/kotlin/org/wfanet/virtualpeople/core/model/utils/PopulationNodeHelper.kt` line 27 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/84#-Os8SMOd4fGGE_9FgaAt:-Os8SMOd4fGGE_9FgaAu:b-rhbe3p) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/e9bcde9e29b13a462bd4d37988bccb44f24bf5a1/src/main/kotlin/org/wfanet/virtualpeople/core/model/utils/PopulationNodeHelper.kt#L27)):*
> ```Kotlin
>  * Keeps seed computation and quantum label collapsing in sync across both node types.
>  */
> object PopulationNodeHelper {
> ```

This PR should also update PopulationNodeImpl to call PopulationNodeHelper, otherwise there are still two copies of the same logic that can drift
  independently

Also - needs unit tests
___
*[`src/test/kotlin/org/wfanet/virtualpeople/core/model/utils/FeistelTest.kt` line 102 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/84#-Os8S0nV5p8kv47MCmXG:-Os8S0nV5p8kv47MCmXH:b-rrczmb) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/e9bcde9e29b13a462bd4d37988bccb44f24bf5a1/src/test/kotlin/org/wfanet/virtualpeople/core/model/utils/FeistelTest.kt#L102)):*
> ```Kotlin
>     // Pin the values so any algorithm change is detected.
>     vectors.forEachIndexed { i, (v, d, s) ->
>       assertEquals(expected[i], Feistel.permute(v, d, s),
> ```

this test is self-referential and will always pass. please fix.
___
*[`src/main/kotlin/org/wfanet/virtualpeople/core/model/utils/Feistel.kt` line 32 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/84#-Os8TM6S3tleE1mU04P5:-Os8TM6S3tleE1mU04P6:biqpro) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/e9bcde9e29b13a462bd4d37988bccb44f24bf5a1/src/main/kotlin/org/wfanet/virtualpeople/core/model/utils/Feistel.kt#L32)):*
> ```Kotlin
>    * Must produce identical output to the C++ FeistelPermute for cross-language determinism.
>    */
>   // Recursive cycle-walking matches the C++ FeistelPermute implementation.
> ```

is this comment correct? there's no c++ impl that I know about. if that's the case, then please make this not recursive, per my earlier comment.
___
*[`src/main/kotlin/org/wfanet/virtualpeople/core/model/RankedPopulationNodeImpl.kt` line 102 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/84#-Os8UFkF4MLSKSlo7gBG:-Os8UFkF4MLSKSlo7gBH:b-47d5w2) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/e9bcde9e29b13a462bd4d37988bccb44f24bf5a1/src/main/kotlin/org/wfanet/virtualpeople/core/model/RankedPopulationNodeImpl.kt#L102)):*
> ```Kotlin
>       val config = nodeConfig.rankedPopulationNode
> 
>       require(config.poolsCount > 0) { "RankedPopulationNode must have at least one pool." }
> ```

really this should be `==1` - see my other comment


<!-- Sent from Reviewable.io -->

stevenwarejones (changes requested):
> @stevenwarejones reviewed 9 files and all commit messages, made 2 comments, and resolved 5 discussions.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/84)* status: all files reviewed, 3 unresolved discussions (waiting on alxlyj and Marco-Premier).

___
*[`MODULE.bazel` line 52 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/84#-OsMra6kDIzxblLJik14:-OsMra6kDIzxblLJik15:b443d5y) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/1c29da62e0c888b41ccce8b43d65359a452fa765/MODULE.bazel#L52)):*
> ```Starlark
> bazel_dep(
>     name = "rules_cc",
>     version = "0.1.1",
> ```

you don't need this change in this PR right? it should be scoped to the C++ one.
___
*[`src/test/kotlin/org/wfanet/virtualpeople/core/labeler/RankedLabelingIntegrationTest.kt` line 187 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/84#-OsMsdFo3kp1sQVFOAzb:-OsMsdFo3kp1sQVFOAzc:b-mdkupp) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/1c29da62e0c888b41ccce8b43d65359a452fa765/src/test/kotlin/org/wfanet/virtualpeople/core/labeler/RankedLabelingIntegrationTest.kt#L187)):*
> ```Kotlin
> 
>   @Test
>   fun `ranked size zero behaves like hash based`() {
> ```

  The test checks that VIDs land in the correct range, but doesn't verify that RankedPopulationNode(ranked_size=0, FULL_POOL) produces identical VIDs to PopulationNode with the same pool config and seed. Add a test that creates both  
  node types side by side, feeds the same events, and asserts VID equality. Something like:
```                                                                                                                                                                                                                                      
  @Test                                                           
  fun `ranked size zero produces same VIDs as PopulationNode`() {
    val rankedRoot = CompiledNode.newBuilder()                                                                                                                                                                                            
    TextFormat.merge("""                                                                                                                                                                                                                  
      name: "RankedNode"                                                                                                                                                                                                                  
      ranked_population_node {                                                                                                                                                                                                            
        pools { population_offset: 100 total_population: 500 }    
        random_seed: "same-seed"                                                                                                                                                                                                          
        ranked_size: 0                                            
        unranked_mode: FULL_POOL                                                                                                                                                                                                          
      }                                                                                                                                                                                                                                   
    """, rankedRoot)
                                                                                                                                                                                                                                          
    val populationRoot = CompiledNode.newBuilder()                                                                                                                                                                                        
    TextFormat.merge("""
      name: "PopNode"                                                                                                                                                                                                                     
      population_node {                                                                                                                                                                                                                   
        pools { population_offset: 100 total_population: 500 }
        random_seed: "same-seed"                                                                                                                                                                                                          
      }                                                           
    """, populationRoot)                                                                                                                                                                                                                  
                                                                  
    val rankedLabeler = Labeler.build(rankedRoot.build())                                                                                                                                                                                 
    val popLabeler = Labeler.build(populationRoot.build())
                                                                                                                                                                                                                                          
    for (i in 0 until 100) {                                                                                                                                                                                                              
      val input = LabelerInput.newBuilder()                                                                                                                                                                                               
        .setEventId(EventId.newBuilder().setPublisher("test").setId("event-$i"))                                                                                                                                                          
        .build()                                                                                                                                                                                                                          
      val rankedVid = rankedLabeler.label(input).peopleList[0].virtualPersonId                                                                                                                                                            
      val popVid = popLabeler.label(input).peopleList[0].virtualPersonId                                                                                                                                                                  
      assertEquals(popVid, rankedVid, "VID mismatch for event-$i")                                                                                                                                                                        
    }                                                                                                                                                                                                                                     
  }                                                               
```                                                                                                                                                                                                                              
  This proves ranked_size=0 is a true drop-in replacement — important for model authorities migrating existing PopulationNode leaves.


<!-- Sent from Reviewable.io -->

stevenwarejones (changes requested):
> @stevenwarejones reviewed 2 files and all commit messages, made 1 comment, and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/84)* status: all files reviewed, 1 unresolved discussion (waiting on alxlyj and Marco-Premier).

___
*[`MODULE.bazel` line 52 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/84#-OsMra6kDIzxblLJik14:-OsWD3-iBv6zj3lZz63c:b-3onbi8) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/1c29da62e0c888b41ccce8b43d65359a452fa765/MODULE.bazel#L52)):*

you don't need this change in this PR right? it should be scoped to the C++ one.
</blockquote></details>

virtual-people-common v0.6.0 declares rules_cc 0.1.0, not 0.1.1. Bazel module resolution will pull in 0.1.0 transitively. If there's a specific reason you need 0.1.1, add a comment explaining why. Otherwise, remove this override — let the transitive dep resolve it.

Marco-Premier (approved):
> @Marco-Premier reviewed all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/84)* status: all files reviewed, 1 unresolved discussion (waiting on alxlyj).

Marco-Premier (approved):
please address comment on the module.bazel file 

> @Marco-Premier made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/84#-:-OsWMW3dAPce8TpCcZDy:b7u44ye)* status: all files reviewed, 1 unresolved discussion (waiting on alxlyj).

stevenwarejones (approved):
> @stevenwarejones reviewed 3 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/84)* status: all files reviewed, 1 unresolved discussion (waiting on alxlyj).

___
*[`MODULE.bazel` line 52 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/84#-OsMra6kDIzxblLJik14:-OsYaVmi6aHtEMPy_Qce:bn0wmfy) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/1c29da62e0c888b41ccce8b43d65359a452fa765/MODULE.bazel#L52)):*

virtual-people-common v0.6.0 declares rules_cc 0.1.0, not 0.1.1. Bazel module resolution will pull in 0.1.0 transitively. If there's a specific reason you need 0.1.1, add a comment explaining why. Otherwise, remove this override — let the transitive dep resolve it.
</blockquote></details>

sgtm - can you fix virtual-people-common at the same time?

stevenwarejones (approved):
> @stevenwarejones reviewed 1 file and all commit messages, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/84)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on alxlyj).

raimundoltdf (approved):
:lgtm:

> @raimundoltdf made 3 comments.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/84#-:-OtF8q1G9WmGz5hkxc6W:bnfp4nl)* status: all files reviewed, 2 unresolved discussions (waiting on alxlyj).

___
*[`src/test/kotlin/org/wfanet/virtualpeople/core/labeler/RankedLabelingIntegrationTest.kt` line 28 at r9](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/84#-OtF7Jg81joRlkw5oUap:-OtF7Jg81joRlkw5oUaq:bpjasf4) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/f2ab15f037e0fa50373739847066c8dddc04b45d/src/test/kotlin/org/wfanet/virtualpeople/core/labeler/RankedLabelingIntegrationTest.kt#L28)):*
> ```Kotlin
> 
> @RunWith(JUnit4::class)
> class RankedLabelingIntegrationTest {
> ```

Thanks for writing these tests. 

Should we add some tests for handling the cases where the event carries multiple rank assignments and make sure they are resolved correctly?

  https://github.com/world-federation-of-advertisers/virtual-people-common/blob/main/src/main/proto/wfa/virtual_people/common/event.proto#L171-L173
___
*[`src/test/kotlin/org/wfanet/virtualpeople/core/labeler/RankedLabelingIntegrationTest.kt` line 167 at r9](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/84#-OtF79M29YL9AS1_0D-8:-OtF79M29YL9AS1_0D-9:b-qqgjt6) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/f2ab15f037e0fa50373739847066c8dddc04b45d/src/test/kotlin/org/wfanet/virtualpeople/core/labeler/RankedLabelingIntegrationTest.kt#L167)):*
> ```Kotlin
>     val labeler = Labeler.build(root.build())
> 
>     // local_rank=250 >= ranked_size=200 — should fall back to unranked DISJOINT range.
> ```

maybe worth testing the boundary condition (local_rank=ranked_size) and make sure you get the correct behavior (avoid off by 1 errors).

alxlyj (commented):
> @alxlyj reviewed 1 file, resolved 2 discussions, and dismissed @raimundoltdf from 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/84)* status: 13 of 14 files reviewed, 1 unresolved discussion (waiting on stevenwarejones).

alxlyj (commented):
> @alxlyj reviewed 1 file and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/84)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on alxlyj).

PR #83, opened 2026-04-15
Add pass-1 (pool-identity) labeling mode

alxlyj:
## Summary

C++ pass-1 (pool-identity) labeling mode for the two-pass memoized VID pipeline. Counterpart to Kotlin (PR #87).

### Changes
- `LabelingMode` enum (kFull, kPoolIdentity) + overloaded `Label(input, output, mode)`
- RankedPopulationNodeImpl emits PoolAssignment in pass-1 mode (checked first in Apply)
- PopulationNodeImpl rejects pass-1 mode (two-pass requires fully-ranked trees)
- Labeler copies pool_assignments to output

Depends on virtual-people-common v0.7.0 (pool_identity_mode field). Stacked on #81, #82.

## Test Plan

- [x] `bazel test //src/test/cc/wfa/virtual_people/core/labeler:pass1_mode_test`
- [x] `bazel test //src/test/cc/wfa/virtual_people/core/labeler:ranked_labeling_integration_test`
  - Two-pass collision-free, branch routing to correct pools, ranked_size=0 fallback, PopulationNode rejects pass-1

PR #87, opened 2026-05-26
Add pass-1 (pool-identity) labeling mode (Kotlin)

alxlyj:
## Summary

Add pass-1 labeling mode for the two-pass memoized VID assignment pipeline.

### Changes

**Labeler.kt:**
- `LabelingMode` enum (`FULL`, `POOL_IDENTITY`)
- Overloaded `label(input, mode)` — sets `pool_identity_mode` on event, copies `pool_assignments` to output

**RankedPopulationNodeImpl.kt:**
- Check `pool_identity_mode` at start of `apply()` — emit `PoolAssignment` and return early
- `PopulationNodeImpl` is unaffected (backward-compatible)

Depends on virtual-people-common PR #73 (`pool_identity_mode` field on `LabelerEvent`).

## Test Plan

- [ ] `bazel test //src/test/kotlin/org/wfanet/virtualpeople/core/labeler:RankedLabelingIntegrationTest`
  - Pass-1 emits pool assignment without VID
  - PopulationNode unaffected in pass-1 mode
  - Full two-pass collision-free end-to-end (100 events)

### DIFF
--- a/src/main/kotlin/org/wfanet/virtualpeople/core/model/PopulationNodeImpl.kt
+++ b/src/main/kotlin/org/wfanet/virtualpeople/core/model/PopulationNodeImpl.kt
@@ -36,11 +36,11 @@ private constructor(
    * assigned to virtual_person_activities[0] in [event].
    */
   override fun apply(event: LabelerEvent.Builder) {
+    // Pass-1 (pool-identity) mode: a plain PopulationNode has no ranked pool to
+    // announce, so it produces no output. This lets a model mix ranked and
+    // unranked leaves; events routing here are labeled normally in pass-2.
     if (event.poolIdentityMode) {
-      error(
-        "PopulationNodeImpl does not support pool-identity mode. " +
-          "Use RankedPopulationNode for two-pass labeling."
-      )
+      return
     }
 
     // Creates a new virtual_person_activities in the event and writes the
--- a/src/main/kotlin/org/wfanet/virtualpeople/core/model/PopulationNodeImpl.kt
+++ b/src/main/kotlin/org/wfanet/virtualpeople/core/model/PopulationNodeImpl.kt
@@ -40,6 +40,13 @@ private constructor(
      * Creates a new virtual_person_activities in [event] and write the virtual person id and label.
      * No virtual_person_activity should be added by previous nodes.
      */
+    if (event.poolIdentityMode) {
+      error(
+        "PopulationNodeImpl does not support pool-identity mode. " +
+          "Use RankedPopulationNode for two-pass labeling."
+      )
+    }
+
     if (event.virtualPersonActivitiesCount > 0) {
       error("virtual_person_activities should only be created in leaf nodes.")
     }
--- a/src/main/kotlin/org/wfanet/virtualpeople/core/model/RankedPopulationNodeImpl.kt
+++ b/src/main/kotlin/org/wfanet/virtualpeople/core/model/RankedPopulationNodeImpl.kt
@@ -16,7 +16,7 @@ package org.wfanet.virtualpeople.core.model
 
 import org.wfanet.virtualpeople.common.CompiledNode
 import org.wfanet.virtualpeople.common.LabelerEvent
-import org.wfanet.virtualpeople.common.PoolAssignment
+import org.wfanet.virtualpeople.common.poolAssignment
 import org.wfanet.virtualpeople.common.RankedPopulationNode.UnrankedMode
 import org.wfanet.virtualpeople.common.VirtualPersonActivity
 import org.wfanet.virtualpeople.core.model.utils.Feistel
@@ -44,12 +44,11 @@ private constructor(
   override fun apply(event: LabelerEvent.Builder) {
     // Pass-1 mode: emit pool identity and return without assigning a VID.
     if (event.poolIdentityMode) {
-      event.addPoolAssignments(
-        PoolAssignment.newBuilder()
-          .setPoolOffset(poolOffset.toLong())
-          .setPoolSize(poolSize.toLong())
-          .setRankedSize(rankedSize.toLong())
-      )
+      event.addPoolAssignments(poolAssignment {
+        poolOffset = this@RankedPopulationNodeImpl.poolOffset.toLong()
+        poolSize = this@RankedPopulationNodeImpl.poolSize.toLong()
+        rankedSize = this@RankedPopulationNodeImpl.rankedSize.toLong()
+      })
       return
     }
 
--- a/src/main/kotlin/org/wfanet/virtualpeople/core/model/RankedPopulationNodeImpl.kt
+++ b/src/main/kotlin/org/wfanet/virtualpeople/core/model/RankedPopulationNodeImpl.kt
@@ -44,11 +44,13 @@ private constructor(
   override fun apply(event: LabelerEvent.Builder) {
     // Pass-1 mode: emit pool identity and return without assigning a VID.
     if (event.poolIdentityMode) {
-      event.addPoolAssignments(poolAssignment {
-        poolOffset = this@RankedPopulationNodeImpl.poolOffset.toLong()
-        poolSize = this@RankedPopulationNodeImpl.poolSize.toLong()
-        rankedSize = this@RankedPopulationNodeImpl.rankedSize.toLong()
-      })
+      event.addPoolAssignments(
+        poolAssignment {
+          poolOffset = this@RankedPopulationNodeImpl.poolOffset.toLong()
+          poolSize = this@RankedPopulationNodeImpl.poolSize.toLong()
+          rankedSize = this@RankedPopulationNodeImpl.rankedSize.toLong()
+        }
+      )
       return
     }
 
--- a/src/main/kotlin/org/wfanet/virtualpeople/core/model/RankedPopulationNodeImpl.kt
+++ b/src/main/kotlin/org/wfanet/virtualpeople/core/model/RankedPopulationNodeImpl.kt
@@ -16,9 +16,9 @@ package org.wfanet.virtualpeople.core.model
 
 import org.wfanet.virtualpeople.common.CompiledNode
 import org.wfanet.virtualpeople.common.LabelerEvent
-import org.wfanet.virtualpeople.common.poolAssignment
 import org.wfanet.virtualpeople.common.RankedPopulationNode.UnrankedMode
 import org.wfanet.virtualpeople.common.VirtualPersonActivity
+import org.wfanet.virtualpeople.common.poolAssignment
 import org.wfanet.virtualpeople.core.model.utils.Feistel
 import org.wfanet.virtualpeople.core.model.utils.PopulationNodeHelper
 import org.wfanet.virtualpeople.core.model.utils.jumpConsistentHash
--- a/src/main/kotlin/org/wfanet/virtualpeople/core/model/RankedPopulationNodeImpl.kt
+++ b/src/main/kotlin/org/wfanet/virtualpeople/core/model/RankedPopulationNodeImpl.kt
@@ -16,8 +16,8 @@ package org.wfanet.virtualpeople.core.model
 
 import org.wfanet.virtualpeople.common.CompiledNode
 import org.wfanet.virtualpeople.common.LabelerEvent
-import org.wfanet.virtualpeople.common.RankedPopulationNode.UnrankedMode
 import org.wfanet.virtualpeople.common.PoolAssignment
+import org.wfanet.virtualpeople.common.RankedPopulationNode.UnrankedMode
 import org.wfanet.virtualpeople.common.VirtualPersonActivity
 import org.wfanet.virtualpeople.core.model.utils.Feistel
 import org.wfanet.virtualpeople.core.model.utils.PopulationNodeHelper
--- a/src/main/kotlin/org/wfanet/virtualpeople/core/model/RankedPopulationNodeImpl.kt
+++ b/src/main/kotlin/org/wfanet/virtualpeople/core/model/RankedPopulationNodeImpl.kt
@@ -17,6 +17,7 @@ package org.wfanet.virtualpeople.core.model
 import org.wfanet.virtualpeople.common.CompiledNode
 import org.wfanet.virtualpeople.common.LabelerEvent
 import org.wfanet.virtualpeople.common.RankedPopulationNode.UnrankedMode
+import org.wfanet.virtualpeople.common.PoolAssignment
 import org.wfanet.virtualpeople.common.VirtualPersonActivity
 import org.wfanet.virtualpeople.core.model.utils.Feistel
 import org.wfanet.virtualpeople.core.model.utils.PopulationNodeHelper
@@ -41,6 +42,17 @@ private constructor(
 ) : ModelNode(nodeConfig) {
 
   override fun apply(event: LabelerEvent.Builder) {
+    // Pass-1 mode: emit pool identity and return without assigning a VID.
+    if (event.poolIdentityMode) {
+      event.addPoolAssignments(
+        PoolAssignment.newBuilder()
+          .setPoolOffset(poolOffset.toLong())
+          .setPoolSize(poolSize.toLong())
+          .setRankedSize(rankedSize.toLong())
+      )
+      return
+    }
+
     if (event.virtualPersonActivitiesCount > 0) {
       error("virtual_person_activities should only be created in leaf nodes.")
     }
--- a/src/main/kotlin/org/wfanet/virtualpeople/core/model/RankedPopulationNodeImpl.kt
+++ b/src/main/kotlin/org/wfanet/virtualpeople/core/model/RankedPopulationNodeImpl.kt
@@ -59,12 +59,16 @@ private constructor(
 
     val activity = VirtualPersonActivity.newBuilder()
 
-    val rankAssignment =
-      if (event.hasLabelerInput()) {
-        event.labelerInput.rankAssignmentsList.firstOrNull { it.poolOffset.toULong() == poolOffset }
-      } else {
-        null
-      }
+    val rankAssignments =
+      if (event.hasLabelerInput()) event.labelerInput.rankAssignmentsList else emptyList()
+    val rankAssignment = rankAssignments.firstOrNull { it.poolOffset.toULong() == poolOffset }
+
+    if (rankAssignments.isNotEmpty() && rankAssignment == null) {
+      error(
+        "RankAssignment provided but none match pool_offset=$poolOffset. " +
+          "Available: ${rankAssignments.map { it.poolOffset }}."
+      )
+    }
 
     val virtualPersonId =
       if (rankAssignment != null && rankAssignment.localRank.toULong() < rankedSize) {
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -33,7 +33,7 @@ bazel_dep(
 )
 bazel_dep(
     name = "virtual-people-common",
-    version = "0.6.1",
+    version = "0.7.0",
     repo_name = "wfa_virtual_people_common",
 )
 
--- a/src/main/kotlin/org/wfanet/virtualpeople/core/labeler/Labeler.kt
+++ b/src/main/kotlin/org/wfanet/virtualpeople/core/labeler/Labeler.kt
@@ -26,18 +26,37 @@ import org.wfanet.virtualpeople.common.labelerOutput
 import org.wfanet.virtualpeople.core.common.Hashing
 import org.wfanet.virtualpeople.core.model.ModelNode
 
+enum class LabelingMode {
+  FULL,
+  POOL_IDENTITY,
+}
+
 class Labeler private constructor(private val rootNode: ModelNode) {
 
   /** Apply the model to generate the labels. Invalid inputs will result in an error. */
-  fun label(input: LabelerInput): LabelerOutput {
-    // Prepare labeler event.
+  fun label(input: LabelerInput): LabelerOutput = label(input, LabelingMode.FULL)
+
+  /**
+   * Apply the model with a specific labeling mode.
+   *
+   * In [LabelingMode.POOL_IDENTITY] mode, RankedPopulationNode leaves emit PoolAssignment entries
+   * instead of assigning VIDs.
+   */
+  fun label(input: LabelerInput, mode: LabelingMode): LabelerOutput {
     val eventBuilder = labelerEvent { labelerInput = input }.toBuilder()
     setFingerprints(eventBuilder)
+
+    if (mode == LabelingMode.POOL_IDENTITY) {
+      eventBuilder.poolIdentityMode = true
+    }
 
     rootNode.apply(eventBuilder)
 
     return labelerOutput {
       people.addAll(eventBuilder.virtualPersonActivitiesList)
+      if (eventBuilder.poolAssignmentsCount > 0) {
+        poolAssignments.addAll(eventBuilder.poolAssignmentsList)
+      }
       serializedDebugTrace = eventBuilder.toString()
     }
   }
--- a/src/main/kotlin/org/wfanet/virtualpeople/core/labeler/Labeler.kt
+++ b/src/main/kotlin/org/wfanet/virtualpeople/core/labeler/Labeler.kt
@@ -53,10 +53,8 @@ class Labeler private constructor(private val rootNode: ModelNode) {
     rootNode.apply(eventBuilder)
 
     return labelerOutput {
-      people.addAll(eventBuilder.virtualPersonActivitiesList)
-      if (eventBuilder.poolAssignmentsCount > 0) {
-        poolAssignments.addAll(eventBuilder.poolAssignmentsList)
-      }
+      people += eventBuilder.virtualPersonActivitiesList
+      poolAssignments += eventBuilder.poolAssignmentsList
       serializedDebugTrace = eventBuilder.toString()
     }
   }
--- a/src/main/kotlin/org/wfanet/virtualpeople/core/model/BranchNodeImpl.kt
+++ b/src/main/kotlin/org/wfanet/virtualpeople/core/model/BranchNodeImpl.kt
@@ -25,19 +25,19 @@ import org.wfanet.virtualpeople.core.model.utils.FieldFiltersMatcher
  * The implementation of the CompiledNode with branch_node set.
  *
  * @param nodeConfig the compiledNode used to build the branch node. The field branch_node must be
- * set.
+ *   set.
  * @param childNodes include the child nodes in all the branches, in the same order as the branches
- * in nodeConfig.
+ *   in nodeConfig.
  * @param hashing if chance is set in each branches in nodeConfig, hashing is set and used to select
- * a child node with randomSeed when Apply is called.
+ *   a child node with randomSeed when Apply is called.
  * @param randomSeed a seed used by the hashing.
  * @param matcher if condition is set in each branches in nodeConfig, matcher is set and used to
- * select the first child node whose condition matches when Apply is called.
+ *   select the first child node whose condition matches when Apply is called.
  * @param updaters if updates is set in nodeConfig, updaters is set. When calling Apply, entries of
- * [updaters] is applied to the event in order. Each entry of updaters updates the value of some
- * fields in the event.
+ *   [updaters] is applied to the event in order. Each entry of updaters updates the value of some
+ *   fields in the event.
  * @param multiplicity if multiplicity is set in nodeConfig, multiplicity is set. When calling
- * Apply, call ApplyMultiplicity.
+ *   Apply, call ApplyMultiplicity.
  */
 internal class BranchNodeImpl
 private constructor(
@@ -99,6 +99,14 @@ private constructor(
       clone.virtualPersonActivitiesList.forEach { person ->
         event.addVirtualPersonActivities(person)
       }
+      /**
+       * Fold back pool assignments too. In pool-identity (pass-1) mode, leaf nodes emit pool
+       * assignments instead of virtual person activities; without this, assignments from cloned
+       * multiplicity events would be dropped.
+       */
+      clone.poolAssignmentsList.forEach { poolAssignment ->
+        event.addPoolAssignments(poolAssignment)
+      }
     }
     return
   }
@@ -129,13 +137,13 @@ private constructor(
      * 1. [nodeConfig].branch_node is not set.
      * 2. [nodeConfig].branch_node.branches is empty.
      * 3. There is at least one of [nodeConfig].branch_node.branches, which has neither node_index
-     * nor node set.
+     *    nor node set.
      * 4. There is at least one of [nodeConfig].branch_node.branches, which has neither chance nor
-     * condition set.
+     *    condition set.
      * 5. At least one of [nodeConfig].branch_node.branches has chance set, and at least one of
-     * [nodeConfig].branch_node.branches has condition set.
+     *    [nodeConfig].branch_node.branches has condition set.
      * 6. When node_index is set in [nodeConfig].branch_node.branches. [nodeRefs] has no entry for
-     * this node_index.
+     *    this node_index.
      *
      * For any [nodeConfig].branch_node.branches with node_index set, the corresponding child node
      * is retrieved from [nodeRefs] using node_index as the key.
--- a/src/test/kotlin/org/wfanet/virtualpeople/core/labeler/RankedLabelingIntegrationTest.kt
+++ b/src/test/kotlin/org/wfanet/virtualpeople/core/labeler/RankedLabelingIntegrationTest.kt
@@ -422,12 +422,13 @@ class RankedLabelingIntegrationTest {
     // Pass 2: inject ranks and assign VIDs.
     val vids = mutableSetOf<ULong>()
     inputs.forEachIndexed { rank, input ->
-      val rankedInput = input.copy {
-        rankAssignments += rankAssignment {
-          poolOffset = 0
-          localRank = rank.toLong()
+      val rankedInput =
+        input.copy {
+          rankAssignments += rankAssignment {
+            poolOffset = 0
+            localRank = rank.toLong()
+          }
         }
-      }
 
       val output = labeler.label(rankedInput)
       assertEquals(1, output.peopleCount)
@@ -462,10 +463,12 @@ class RankedLabelingIntegrationTest {
         val rankedInput =
           input
             .toBuilder()
-            .addRankAssignments(rankAssignment {
-              poolOffset = 0
-              localRank = rank.toLong()
-            })
+            .addRankAssignments(
+              rankAssignment {
+                poolOffset = 0
+                localRank = rank.toLong()
+              }
+            )
             .build()
         labeler.label(rankedInput).peopleList[0].virtualPersonId
       }
--- a/src/test/kotlin/org/wfanet/virtualpeople/core/labeler/RankedLabelingIntegrationTest.kt
+++ b/src/test/kotlin/org/wfanet/virtualpeople/core/labeler/RankedLabelingIntegrationTest.kt
@@ -15,6 +15,7 @@
 package org.wfanet.virtualpeople.core.labeler
 
 import com.google.protobuf.TextFormat
+import kotlin.test.assertFailsWith
 import kotlin.test.assertEquals
 import kotlin.test.assertTrue
 import org.junit.Test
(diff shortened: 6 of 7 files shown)


stevenwarejones (changes requested):
PopulationNodeImpl doesn't handle poolIdentityMode. If an event routes to a PopulationNode leaf during pass-1, it silently assigns a VID instead of emitting a pool assignment. Two options:

  1. Emit a PoolAssignment from PopulationNodeImpl too (so all leaf types participate in pass-1 uniformly).
  2. Throw an IllegalStateException if poolIdentityMode is true in PopulationNodeImpl.apply() — this makes it explicit that two-pass mode only works with RankedPopulationNode trees, and catches misuse early.

  Option 2 seems safer — if the model tree isn't fully ranked, the caller should know immediately rather than getting mixed results.

> @stevenwarejones reviewed all commit messages and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/87#-:-Ot_tXXM5f_uGW8WhbKg:b-ze3pxw)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on alxlyj, Marco-Premier, and raimundoltdf).

___
*[`src/test/kotlin/org/wfanet/virtualpeople/core/labeler/RankedLabelingIntegrationTest.kt` line 409 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/87#-Ot_sOPL141A4l57GfK5:-Ot_sOPL141A4l57GfK6:b-wydqxo) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/9efa96a995d315c2551a1c760da5027159c130bc/src/test/kotlin/org/wfanet/virtualpeople/core/labeler/RankedLabelingIntegrationTest.kt#L409)):*
> ```Kotlin
>       val output = labeler.label(input, LabelingMode.POOL_IDENTITY)
>       assertEquals(1, output.poolAssignmentsCount)
>       assertEquals(0, output.poolAssignmentsCount.let { output.peopleCount })
> ```

do you just mean
```
assertEquals(0, output.peopleCount)
```
or do you mean
```
  assertEquals(0, output.poolAssignmentsCount)
  assertEquals(0, output.peopleCount)
```

stevenwarejones (changes requested):
> @stevenwarejones made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/87)* status: 0 of 3 files reviewed, 2 unresolved discussions (waiting on alxlyj, Marco-Premier, and raimundoltdf).

___
*[`src/main/kotlin/org/wfanet/virtualpeople/core/model/RankedPopulationNodeImpl.kt` line 44 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/87#-Ot_uqdwA-Wmce4dlt7Q:-Ot_uqdx63LKd9yLhB9s:b7uufng) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/9efa96a995d315c2551a1c760da5027159c130bc/src/main/kotlin/org/wfanet/virtualpeople/core/model/RankedPopulationNodeImpl.kt#L44)):*
> ```Kotlin
> ) : ModelNode(nodeConfig) {
> 
>   override fun apply(event: LabelerEvent.Builder) {
> ```

1. The pool assignment logic should have unit test coverage in RankedPopulationNodeImplTest — specifically a test with multiple pools verifying that inputs are routed to the correct pool based on their attributes.

2.   There's no test for what happens during pass-2 when the injected rank assignment is invalid — e.g., a rank that's out of bounds for the pool size, or a rank for a pool that doesn't exist. The code should handle this gracefully
  (throw an exception or return an error), and a test should verify that behavior.

stevenwarejones (changes requested):
> @stevenwarejones reviewed all commit messages and made 6 comments.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/87)* status: 0 of 4 files reviewed, 8 unresolved discussions (waiting on alxlyj, Marco-Premier, and raimundoltdf).

___
*[`src/main/kotlin/org/wfanet/virtualpeople/core/labeler/Labeler.kt` line 57 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/87#-Ou2BsYS2hrTD15RfotC:-Ou2BsYS2hrTD15RfotD:b-3lv2s2) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/2f30464a4b4d9d04ed34f7502e83ea317515a658/src/main/kotlin/org/wfanet/virtualpeople/core/labeler/Labeler.kt#L57)):*
> ```Kotlin
>     return labelerOutput {
>       people.addAll(eventBuilder.virtualPersonActivitiesList)
>       if (eventBuilder.poolAssignmentsCount > 0) {
> ```

nit: you donn't really need to check this do you?
___
*[`src/main/kotlin/org/wfanet/virtualpeople/core/labeler/Labeler.kt` line 58 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/87#-Ou2Bppd5WegOIiWQSV8:-Ou2Bppd5WegOIiWQSV9:bxy9mbr) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/2f30464a4b4d9d04ed34f7502e83ea317515a658/src/main/kotlin/org/wfanet/virtualpeople/core/labeler/Labeler.kt#L58)):*
> ```Kotlin
>       people.addAll(eventBuilder.virtualPersonActivitiesList)
>       if (eventBuilder.poolAssignmentsCount > 0) {
>         poolAssignments.addAll(eventBuilder.poolAssignmentsList)
> ```

nit: use the kotlin ksl.
___
*[`src/test/kotlin/org/wfanet/virtualpeople/core/labeler/RankedLabelingIntegrationTest.kt` line 29 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/87#-Ou2DX6iAtxGXqOPLoik:-Ou2DX6iAtxGXqOPLoil:b-uqjn3s) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/2f30464a4b4d9d04ed34f7502e83ea317515a658/src/test/kotlin/org/wfanet/virtualpeople/core/labeler/RankedLabelingIntegrationTest.kt#L29)):*
> ```Kotlin
> 
> @RunWith(JUnit4::class)
> class RankedLabelingIntegrationTest {
> ```

1.   Add a test where localRank equals rankedSize (e.g., rank 200 with rankedSize 200). This is the boundary where ranked assignment should NOT apply and the event should fall back to unranked. The existing "non-existent pool" test covers a mismatched
  poolOffset, but doesn't cover the rank-out-of-bounds edge case — which is a different code path (rankAssignment != null but localRank >= rankedSize).

2.     Add a pass-1 test with a branch node routing to two different RankedPopulationNode leaves with different pool offsets. Verify that each event's PoolAssignment reflects the pool it was routed to based on its attributes. The current pass-1 test uses a 
    single node, so it doesn't verify that pool identity actually varies with routing — which is the main thing pass-1 needs to get right. 
___
*[`src/test/kotlin/org/wfanet/virtualpeople/core/labeler/RankedLabelingIntegrationTest.kt` line 375 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/87#-Ou2CMWUERjnaljX-Dwq:-Ou2CMWUERjnaljX-Dwr:bubafwt) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/2f30464a4b4d9d04ed34f7502e83ea317515a658/src/test/kotlin/org/wfanet/virtualpeople/core/labeler/RankedLabelingIntegrationTest.kt#L375)):*
> ```Kotlin
> 
>   @Test
>   fun `two pass collision free end to end`() {
> ```

can you verify that the 2 pass ran 2x yields the same result. new test.
___
*[`src/test/kotlin/org/wfanet/virtualpeople/core/labeler/RankedLabelingIntegrationTest.kt` line 434 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/87#-Ou2Fh5cBpPGNfrgiBFS:-Ou2Fh5cBpPGNfrgiBFT:b-vzf6c1) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/2f30464a4b4d9d04ed34f7502e83ea317515a658/src/test/kotlin/org/wfanet/virtualpeople/core/labeler/RankedLabelingIntegrationTest.kt#L434)):*
> ```Kotlin
> 
>   @Test
>   fun `rank for non-existent pool falls back to unranked`() {
> ```

If a RankAssignment doesn't match any pool the event routes to, that's caller misuse — it should throw, not silently produce an unranked VID. The
  test should expect an exception instead of asserting fallback range.
___
*[`src/test/kotlin/org/wfanet/virtualpeople/core/labeler/RankedLabelingIntegrationTest.kt` line 453 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/87#-Ou2CrtmDKzdMTCULVQz:-Ou2CrtmDKzdMTCULVR-:bikdfxd) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/2f30464a4b4d9d04ed34f7502e83ea317515a658/src/test/kotlin/org/wfanet/virtualpeople/core/labeler/RankedLabelingIntegrationTest.kt#L453)):*
> ```Kotlin
>     // Rank assignment for pool_offset=9999 which doesn't match pool_offset=100.
>     val input =
>       LabelerInput.newBuilder()
> ```

nit use the kotlin dsls all across this PR

stevenwarejones (changes requested):
> @stevenwarejones reviewed 4 files and all commit messages, made 2 comments, and resolved 8 discussions.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/87)* status: all files reviewed, 2 unresolved discussions (waiting on alxlyj, Marco-Premier, and raimundoltdf).

___
*[`src/main/kotlin/org/wfanet/virtualpeople/core/model/RankedPopulationNodeImpl.kt` line 48 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/87#-Ou3wXtYDaivrTryT1l4:-Ou3wXtYDaivrTryT1l5:b-iwzpqf) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/a5fd6d37da75d738385c1977576e71c84cbd26c1/src/main/kotlin/org/wfanet/virtualpeople/core/model/RankedPopulationNodeImpl.kt#L48)):*
> ```Kotlin
>     if (event.poolIdentityMode) {
>       event.addPoolAssignments(
>         PoolAssignment.newBuilder()
> ```

nit: use kotlin dsls - please use them everyehwere
___
*[`src/test/kotlin/org/wfanet/virtualpeople/core/labeler/RankedLabelingIntegrationTest.kt` line 201 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/87#-Ou3x1ZO6sBOa6ch2K6I:-Ou3x1ZO6sBOa6ch2K6J:b-71xd95) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/a5fd6d37da75d738385c1977576e71c84cbd26c1/src/test/kotlin/org/wfanet/virtualpeople/core/labeler/RankedLabelingIntegrationTest.kt#L201)):*
> ```Kotlin
> 
>   @Test
>   fun `multiple rank assignments resolves correct pool`() {
> ```

  This test used to send rank assignments for multiple pools (100, 500, 9000) to verify the node picks the one matching its pool_offset. Now it only sends the matching one (500), so it's just testing the single-assignment case again. Restore the other
  assignments — the throw logic only fires when none match, so having extra non-matching assignments alongside the correct one should still work and is what this test should be covering.

stevenwarejones (commented):
> @stevenwarejones made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/87)* status: all files reviewed, 3 unresolved discussions (waiting on alxlyj, Marco-Premier, and raimundoltdf).

___
*[`src/test/kotlin/org/wfanet/virtualpeople/core/labeler/RankedLabelingIntegrationTest.kt` line 455 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/87#-Ou3xPce3qUFyFulmmSb:-Ou3xPce3qUFyFulmmSc:bwyqnr) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/a5fd6d37da75d738385c1977576e71c84cbd26c1/src/test/kotlin/org/wfanet/virtualpeople/core/labeler/RankedLabelingIntegrationTest.kt#L455)):*
> ```Kotlin
>       return inputs.mapIndexed { rank, input ->
>         val rankedInput =
>           input
> ```

nit: this is more kotlin-esque
```
  val rankedInput = input.copy {
    rankAssignments += rankAssignment {
      poolOffset = 0
      localRank = rank.toLong()
    }
  }
```

stevenwarejones (commented):
Can you resolve the conflicts? Also, I don't believe it builds currently.

> @stevenwarejones reviewed all commit messages, made 3 comments, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/87#-:-OuHsQQ49c3A1XursJYb:b-ubrml7)* status: 2 of 4 files reviewed, 3 unresolved discussions (waiting on alxlyj, Marco-Premier, and raimundoltdf).

___
*[`src/test/kotlin/org/wfanet/virtualpeople/core/labeler/RankedLabelingIntegrationTest.kt` line 455 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/87#-Ou3xPce3qUFyFulmmSb:-OuHqhFH5MdWHnpUzw-1:b42ok4p) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/a5fd6d37da75d738385c1977576e71c84cbd26c1/src/test/kotlin/org/wfanet/virtualpeople/core/labeler/RankedLabelingIntegrationTest.kt#L455)):*

nit: this is more kotlin-esque
```
  val rankedInput = input.copy {
    rankAssignments += rankAssignment {
      poolOffset = 0
      localRank = rank.toLong()
    }
  }
```

</blockquote></details>

copy {} is available for LabelerInput — it's generated as a top-level extension function by protoc-gen-kotlin. You just need to add import org.wfanet.virtualpeople.common.copy. Then replace the toBuilder().addRankAssignments(...).build() calls
  with input.copy { rankAssignments += rankAssignment { ... } }. I verified this builds and all tests pass on bazel-server.
___
*[`src/main/kotlin/org/wfanet/virtualpeople/core/model/PopulationNodeImpl.kt` line 40 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/87#-OuHqQwE-uzEc_aEDgcU:-OuHqQwE-uzEc_aEDgcV:bcxapoj) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/e3d974517bbcd5ef76f7b65fea76a2e89ad5c35a/src/main/kotlin/org/wfanet/virtualpeople/core/model/PopulationNodeImpl.kt#L40)):*
> ```Kotlin
>   override fun apply(event: LabelerEvent.Builder) {
>     /**
>      * Creates a new virtual_person_activities in [event] and write the virtual person id and label.
> ```

Nit: The block comment "Creates a new virtual_person_activities..." now sits between the method opening brace and the poolIdentityMode guard, which makes it read as if it describes the guard clause. Move the guard above the comment, or convert
  the comment to a regular // comment since the method-level KDoc already covers the contract.)

alxlyj (commented):
> @alxlyj made 12 comments and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/87)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on Marco-Premier, raimundoltdf, and stevenwarejones).

___
*[`src/main/kotlin/org/wfanet/virtualpeople/core/labeler/Labeler.kt` line 57 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/87#-Ou2BsYS2hrTD15RfotC:-OuI8vSk902lpnT8if5x:b-b8tqvk) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/2f30464a4b4d9d04ed34f7502e83ea317515a658/src/main/kotlin/org/wfanet/virtualpeople/core/labeler/Labeler.kt#L57)):*

nit: you donn't really need to check this do you?
</blockquote></details>

Done. Removed the if (count > 0) guard.
___
*[`src/main/kotlin/org/wfanet/virtualpeople/core/labeler/Labeler.kt` line 58 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/87#-Ou2Bppd5WegOIiWQSV8:-OuI8rx85Ar0vfdWEtRP:bx9nc4n) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/2f30464a4b4d9d04ed34f7502e83ea317515a658/src/main/kotlin/org/wfanet/virtualpeople/core/labeler/Labeler.kt#L58)):*

nit: use the kotlin ksl.
</blockquote></details>

Done. people += and poolAssignments +=.
___
*[`src/main/kotlin/org/wfanet/virtualpeople/core/model/PopulationNodeImpl.kt` line 40 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/87#-OuHqQwE-uzEc_aEDgcU:-OuI3gIQ17D1MZ0FO_Ld:b-l2e1lw) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/e3d974517bbcd5ef76f7b65fea76a2e89ad5c35a/src/main/kotlin/org/wfanet/virtualpeople/core/model/PopulationNodeImpl.kt#L40)):*

Nit: The block comment "Creates a new virtual_person_activities..." now sits between the method opening brace and the poolIdentityMode guard, which makes it read as if it describes the guard clause. Move the guard above the comment, or convert
  the comment to a regular // comment since the method-level KDoc already covers the contract.)
</blockquote></details>

Done. Guard is now the first statement in apply(). Converted block comment to // style.
___
*[`src/main/kotlin/org/wfanet/virtualpeople/core/model/RankedPopulationNodeImpl.kt` line 44 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/87#-Ot_uqdwA-Wmce4dlt7Q:-OuI9-3f8IUFksUC48aa:b-zgxrwo) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/9efa96a995d315c2551a1c760da5027159c130bc/src/main/kotlin/org/wfanet/virtualpeople/core/model/RankedPopulationNodeImpl.kt#L44)):*

1. The pool assignment logic should have unit test coverage in RankedPopulationNodeImplTest — specifically a test with multiple pools verifying that inputs are routed to the correct pool based on their attributes.

2.   There's no test for what happens during pass-2 when the injected rank assignment is invalid — e.g., a rank that's out of bounds for the pool size, or a rank for a pool that doesn't exist. The code should handle this gracefully
  (throw an exception or return an error), and a test should verify that behavior.
</blockquote></details>

Done. Added boundary test (rank == rankedSize → unranked), multiple rank assignments test (pools 100/500/9000), and non-existent pool test (throws).

Rank out of bounds (rank >= rankedSize) falls back to unranked — covered by rank overflow and boundary rank tests. Non-existent pool (no matching pool_offset) throws IllegalStateException — covered by rank for non-existent pool throws test.
___
*[`src/main/kotlin/org/wfanet/virtualpeople/core/model/RankedPopulationNodeImpl.kt` line 48 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/87#-Ou3wXtYDaivrTryT1l4:-OuI4R0C0KXYynx8H8DM:bdstwbx) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/a5fd6d37da75d738385c1977576e71c84cbd26c1/src/main/kotlin/org/wfanet/virtualpeople/core/model/RankedPopulationNodeImpl.kt#L48)):*

nit: use kotlin dsls - please use them everyehwere
</blockquote></details>

Done. Using poolAssignment {} DSL.
___
*[`src/test/kotlin/org/wfanet/virtualpeople/core/labeler/RankedLabelingIntegrationTest.kt` line 409 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/87#-Ot_sOPL141A4l57GfK5:-OuI9agt9DA-Qa8x2yv0:b-l4v90x) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/9efa96a995d315c2551a1c760da5027159c130bc/src/test/kotlin/org/wfanet/virtualpeople/core/labeler/RankedLabelingIntegrationTest.kt#L409)):*

do you just mean
```
assertEquals(0, output.peopleCount)
```
or do you mean
```
  assertEquals(0, output.poolAssignmentsCount)
  assertEquals(0, output.peopleCount)
```
</blockquote></details>

Done. Simplified to assertEquals(0, output.peopleCount).
___
*[`src/test/kotlin/org/wfanet/virtualpeople/core/labeler/RankedLabelingIntegrationTest.kt` line 29 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/87#-Ou2DX6iAtxGXqOPLoik:-OuI9q744Gt9GBsGKJQw:b-oi57ac) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/2f30464a4b4d9d04ed34f7502e83ea317515a658/src/test/kotlin/org/wfanet/virtualpeople/core/labeler/RankedLabelingIntegrationTest.kt#L29)):*

1.   Add a test where localRank equals rankedSize (e.g., rank 200 with rankedSize 200). This is the boundary where ranked assignment should NOT apply and the event should fall back to unranked. The existing "non-existent pool" test covers a mismatched
  poolOffset, but doesn't cover the rank-out-of-bounds edge case — which is a different code path (rankAssignment != null but localRank >= rankedSize).

2.     Add a pass-1 test with a branch node routing to two different RankedPopulationNode leaves with different pool offsets. Verify that each event's PoolAssignment reflects the pool it was routed to based on its attributes. The current pass-1 test uses a 
    single node, so it doesn't verify that pool identity actually varies with routing — which is the main thing pass-1 needs to get right.
</blockquote></details>

Done. Added boundary rank equals ranked size and pass 1 branch node routes to correct pools tests.
___
*[`src/test/kotlin/org/wfanet/virtualpeople/core/labeler/RankedLabelingIntegrationTest.kt` line 375 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/87#-Ou2CMWUERjnaljX-Dwq:-OuI9g8VBzc5fHkWHdGb:bggkyk2) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/2f30464a4b4d9d04ed34f7502e83ea317515a658/src/test/kotlin/org/wfanet/virtualpeople/core/labeler/RankedLabelingIntegrationTest.kt#L375)):*

can you verify that the 2 pass ran 2x yields the same result. new test.
</blockquote></details>

Done. Added two pass is deterministic test — runs twice, asserts identical VIDs.
___
*[`src/test/kotlin/org/wfanet/virtualpeople/core/labeler/RankedLabelingIntegrationTest.kt` line 434 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/87#-Ou2Fh5cBpPGNfrgiBFS:-OuI9u8y45VVij48Wzv8:b-coxlqg) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/2f30464a4b4d9d04ed34f7502e83ea317515a658/src/test/kotlin/org/wfanet/virtualpeople/core/labeler/RankedLabelingIntegrationTest.kt#L434)):*

If a RankAssignment doesn't match any pool the event routes to, that's caller misuse — it should throw, not silently produce an unranked VID. The
  test should expect an exception instead of asserting fallback range.
</blockquote></details>

Done. Throws IllegalStateException when rank assignments are non-empty but none match. Test expects exception.
___
*[`src/test/kotlin/org/wfanet/virtualpeople/core/labeler/RankedLabelingIntegrationTest.kt` line 453 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/87#-Ou2CrtmDKzdMTCULVQz:-OuI9h8_DDBFk1BolW1p:b-z1xp25) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/2f30464a4b4d9d04ed34f7502e83ea317515a658/src/test/kotlin/org/wfanet/virtualpeople/core/labeler/RankedLabelingIntegrationTest.kt#L453)):*

nit use the kotlin dsls all across this PR
</blockquote></details>

Done. All tests use labelerInput {}, eventId {}, rankAssignment {} DSLs.
___
*[`src/test/kotlin/org/wfanet/virtualpeople/core/labeler/RankedLabelingIntegrationTest.kt` line 201 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/87#-Ou3x1ZO6sBOa6ch2K6I:-OuI7iex5JjbSDISztyg:b-yvmclf) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/a5fd6d37da75d738385c1977576e71c84cbd26c1/src/test/kotlin/org/wfanet/virtualpeople/core/labeler/RankedLabelingIntegrationTest.kt#L201)):*

This test used to send rank assignments for multiple pools (100, 500, 9000) to verify the node picks the one matching its pool_offset. Now it only sends the matching one (500), so it's just testing the single-assignment case again. Restore the other
  assignments — the throw logic only fires when none match, so having extra non-matching assignments alongside the correct one should still work and is what this test should be covering.
</blockquote></details>

 Done. Restored — sends assignments for pools 100, 500, and 9000. Only 500 matches the node.
___
*[`src/test/kotlin/org/wfanet/virtualpeople/core/labeler/RankedLabelingIntegrationTest.kt` line 455 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/87#-Ou3xPce3qUFyFulmmSb:-OuIA11LEeqXjTZPOSXg:b93ebw5) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/a5fd6d37da75d738385c1977576e71c84cbd26c1/src/test/kotlin/org/wfanet/virtualpeople/core/labeler/RankedLabelingIntegrationTest.kt#L455)):*

copy {} is available for LabelerInput — it's generated as a top-level extension function by protoc-gen-kotlin. You just need to add import org.wfanet.virtualpeople.common.copy. Then replace the toBuilder().addRankAssignments(...).build() calls
  with input.copy { rankAssignments += rankAssignment { ... } }. I verified this builds and all tests pass on bazel-server.
</blockquote></details>

Done. Using input.copy { rankAssignments += ... }. Added the import, using input.copy {}.

alxlyj (commented):
> @alxlyj reviewed 4 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/87)* status: all files reviewed, 1 unresolved discussion (waiting on Marco-Premier, raimundoltdf, and stevenwarejones).

stevenwarejones (approved):
> @stevenwarejones reviewed 4 files and all commit messages, made 1 comment, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/87)* status: all files reviewed, 1 unresolved discussion (waiting on alxlyj, Marco-Premier, and raimundoltdf).

___
*[`src/test/kotlin/org/wfanet/virtualpeople/core/labeler/RankedLabelingIntegrationTest.kt` line 465 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/87#-OuIAfuH0ZkVfmX-9xzp:-OuIAfuH0ZkVfmX-9xzq:b-yjqmbc) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/089ee23cc96e184b57d81d1e4ea022a2d3f64d1f/src/test/kotlin/org/wfanet/virtualpeople/core/labeler/RankedLabelingIntegrationTest.kt#L465)):*
> ```Kotlin
>         val rankedInput =
>           input
>             .toBuilder()
> ```

nit: use copy

stevenwarejones (approved):
:lgtm: 

> @stevenwarejones made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/87#-:-OuIDRRcATw5LHFjJ_kE:bg7igkv)* status: all files reviewed, 1 unresolved discussion (waiting on alxlyj, Marco-Premier, and raimundoltdf).

alxlyj (commented):
> @alxlyj reviewed 2 files and all commit messages, made 1 comment, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/87)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on Marco-Premier and raimundoltdf).

___
*[`src/test/kotlin/org/wfanet/virtualpeople/core/labeler/RankedLabelingIntegrationTest.kt` line 465 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/87#-OuIAfuH0ZkVfmX-9xzp:-OuIE9hq2GymO3O2W3I7:bgbamra) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/089ee23cc96e184b57d81d1e4ea022a2d3f64d1f/src/test/kotlin/org/wfanet/virtualpeople/core/labeler/RankedLabelingIntegrationTest.kt#L465)):*

nit: use copy
</blockquote></details>

Done. Switched to input.copy {}.

stevenwarejones (approved):
> @stevenwarejones reviewed all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/87)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on Marco-Premier and raimundoltdf).

Marco-Premier (approved):
> @Marco-Premier reviewed all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/87)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on raimundoltdf).

src/main/kotlin/org/wfanet/virtualpeople/core/model/PopulationNodeImpl.kt
```diff
@@ -36,10 +36,16 @@ private constructor(
    * assigned to virtual_person_activities[0] in [event].
    */
   override fun apply(event: LabelerEvent.Builder) {
-    /**
-     * Creates a new virtual_person_activities in [event] and write the virtual person id and label.
-     * No virtual_person_activity should be added by previous nodes.
-     */
+    if (event.poolIdentityMode) {
```
raimundoltdf, line 42: why is this not supported? I guess you could have some mix of non-ranked population nodes and vanilla nodes in the model... if for some reason some ID types/segments of the model do not require ranked approaches.

src/main/kotlin/org/wfanet/virtualpeople/core/model/RankedPopulationNodeImpl.kt
```diff
@@ -41,18 +42,34 @@ private constructor(
 ) : ModelNode(nodeConfig) {
 
   override fun apply(event: LabelerEvent.Builder) {
+    // Pass-1 mode: emit pool identity and return without assigning a VID.
+    if (event.poolIdentityMode) {
```
raimundoltdf, line 46: applyMultiplicity in BranchNodeImpl.kt:97-101 folds back clones' virtualPersonActivitiesList but not poolAssignmentsList, so pass-1 pool assignments get dropped for multiplicity events. Please either clone the pool assignments back too, or throw if multiplicity isn't supported in pass-1. I lean toward the former.

src/main/kotlin/org/wfanet/virtualpeople/core/model/PopulationNodeImpl.kt
```diff
@@ -36,10 +36,16 @@ private constructor(
...
+      error(
```
stevenwarejones: Per the discussion with raimundoltdf — this should no-op (return without emitting a PoolAssignment or VID) rather than throw. A model can mix RankedPopulationNode and PopulationNode leaves. Events routing to PopulationNode during pass-1 just produce no output; pass-2 handles them with normal hash-based assignment. Same change needed on C++ PR #83.

Once changed, the existing PopulationNodeRejectsPass1Mode test needs to verify the no-op instead of expecting an exception. Also add an integration test with a mixed model: BranchNode routing 50/50 to a RankedPopulationNode and a PopulationNode. Run pass-1 — events routing to the ranked leaf emit PoolAssignments, events routing to PopulationNode produce nothing. Verify total PoolAssignment count is roughly half and no VIDs are assigned.

src/main/kotlin/org/wfanet/virtualpeople/core/model/PopulationNodeImpl.kt
stevenwarejones: You're right. It should just no-op here, right?

stevenwarejones (commented):
> @stevenwarejones reviewed 3 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/87)* status: all files reviewed, 3 unresolved discussions (waiting on alxlyj and raimundoltdf).

src/main/kotlin/org/wfanet/virtualpeople/core/model/PopulationNodeImpl.kt
alxlyj: Done.  In pass-1 a plain PopulationNode now no-ops (returns without emitting a PoolAssignment or VID), so a model can mix ranked and vanilla leaves. Events routing to a PopulationNode just produce no pass-1 output and get labeled normally in pass-2. Done in both Kotlin (#87) and C++ (#83).

src/main/kotlin/org/wfanet/virtualpeople/core/model/PopulationNodeImpl.kt
alxlyj: Done, all three:
1. PopulationNode no-ops in pass-1 (returns without emitting a PoolAssignment or VID) instead of throwing.
2. Updated the pass-1 PopulationNode test to verify the no-op (succeeds with 0 people and 0 pool assignments) instead of expecting an error.
3. Added a mixed-model integration test: a BranchNode routing 50/50 to a RankedPopulationNode and a PopulationNode, run in pass-1 — events to the ranked leaf emit a PoolAssignment (offset 1000), events to the PopulationNode emit nothing, no VIDs are assigned, and the two counts sum to the event total.
Same change applied to C++ PR #83.

src/main/kotlin/org/wfanet/virtualpeople/core/model/RankedPopulationNodeImpl.kt
alxlyj: Done. Fixed by folding back poolAssignmentsList from the clones too (your preferred option), in BranchNodeImpl.applyMultiplicity. Added a regression test (multiplicity branch with a RankedPopulationNode child in pass-1: the pass-1 pool-assignment total equals the full-mode virtual-person total; verified it fails without the fold-back). Same fix + test applied to the C++ ApplyMultiplicity for parity.

alxlyj (commented):
> @alxlyj reviewed 4 files and all commit messages, and made 3 comments.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/87)* status: all files reviewed, 3 unresolved discussions (waiting on raimundoltdf and stevenwarejones).

raimundoltdf (approved):

stevenwarejones (approved):
> @stevenwarejones reviewed all commit messages and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/87)* status: all files reviewed, 2 unresolved discussions (waiting on raimundoltdf).

alxlyj (commented):
> @alxlyj resolved 2 discussions and dismissed @raimundoltdf from 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/87)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on alxlyj).

PR #89, opened 2026-06-24
fix: RankedPopulationNode falls back to hash on non-matching rank_assignments

stevenwarejones:
When this leaf's `pool_offset` matches no entry in `LabelerInput.rank_assignments`, the leaf must fall back to the hash path instead of failing the job. The dominant case is Phase-1 overflow: when this subpool reached `ranked_size` and this fingerprint was one of the unranked surplus, the subpool has no rank for it — the design (§ Retention Rule, line 278) explicitly requires "overflow fps are not ranked and fall back to the unranked path at Phase 2."

The labeler tree is deterministic, so each impression routes to exactly one subpool leaf. The `rank_assignments` list can hold multiple entries only because the memoized Phase-2 sink conservatively attaches a `RankAssignment` for every subpool the fingerprint is currently ranked in across this `(DataProvider, ModelLine)`; the leaf then picks its own. Today the leaf throws `InvalidArgument` whenever none matches — failing the whole `VidLabelingJob` on every overflowed fingerprint. This PR aligns the code with the proto KDoc on `LabelerInput.RankAssignment`, which already specifies: "If no RankAssignment matches, the event gets hash-based VID assignment."

In the same PR, also stamps `VirtualPersonActivity.memoized_rank_fallback` (new in [virtual-people-common#75](https://github.com/world-federation-of-advertisers/virtual-people-common/pull/75)) whenever the leaf takes the hash path despite a non-empty `rank_assignments`. Without this signal, the memoized Phase-2 pipeline silently degrades to hash-based VID assignment when subpools saturate, with no operational visibility. The consumer (`VidLabelingSink` in [cross-media-measurement#4072](https://github.com/world-federation-of-advertisers/cross-media-measurement/pull/4072)) aggregates this per `(data provider, model line)` into an OpenTelemetry counter to alarm on a rising fallback rate — the indicator that a subpool's `ranked_size` should be raised.

C++ and Kotlin implementations updated symmetrically. Tests cover all four cases: ranked path (no fallback), no `rank_assignments` (no fallback, no memoization attempted), non-matching `rank_assignments` (fallback set), matching with `local_rank >= ranked_size` (fallback set). Full repo test sweep passes (45/45).

Depends on `virtual-people-common` 0.8.0 (adds `VirtualPersonActivity.memoized_rank_fallback`), which has landed in the WFA Bazel registry; `MODULE.bazel` now pins that release instead of a `git_override`.

Issue: world-federation-of-advertisers/cross-media-measurement#4073

### DIFF
--- a/src/test/cc/wfa/virtual_people/core/model/ranked_population_node_impl_test.cc
+++ b/src/test/cc/wfa/virtual_people/core/model/ranked_population_node_impl_test.cc
@@ -422,11 +422,14 @@ TEST(RankedPopulationNodeImplTest, NoMatchingRankAssignmentFallsBackToHash) {
   ra2->set_local_rank(17);
 
   EXPECT_THAT(node->Apply(event), IsOk());
-  uint64_t vid = event.virtual_person_activities(0).virtual_person_id();
+  const auto& activity = event.virtual_person_activities(0);
+  uint64_t vid = activity.virtual_person_id();
   // Must land in the unranked sub-range [pool_offset + ranked_size,
   // pool_offset + pool_size) = [300, 600) under DISJOINT mode.
   EXPECT_GE(vid, 300);
   EXPECT_LT(vid, 600);
+  // The whole point of this path is to surface that the leaf had to hash-fall-back.
+  EXPECT_TRUE(activity.memoized_rank_fallback());
 }
 
 // The non-matching-rank_assignments path must produce the same VID as the
@@ -458,7 +461,11 @@ TEST(RankedPopulationNodeImplTest,
   event_a.set_acting_fingerprint(42);
   event_a.mutable_labeler_input();
   EXPECT_THAT(node->Apply(event_a), IsOk());
-  uint64_t vid_a = event_a.virtual_person_activities(0).virtual_person_id();
+  const auto& activity_a = event_a.virtual_person_activities(0);
+  uint64_t vid_a = activity_a.virtual_person_id();
+  // No rank_assignments at all => caller did not attempt memoization, so the
+  // fallback signal MUST NOT fire.
+  EXPECT_FALSE(activity_a.memoized_rank_fallback());
 
   // Event B: same fingerprint, rank_assignments only for other pools.
   LabelerEvent event_b;
@@ -467,7 +474,11 @@ TEST(RankedPopulationNodeImplTest,
   ra->set_pool_offset(9000);
   ra->set_local_rank(5);
   EXPECT_THAT(node->Apply(event_b), IsOk());
-  uint64_t vid_b = event_b.virtual_person_activities(0).virtual_person_id();
+  const auto& activity_b = event_b.virtual_person_activities(0);
+  uint64_t vid_b = activity_b.virtual_person_id();
+  // Non-matching assignments => caller did attempt memoization but the leaf
+  // hash-fell-back, so the signal MUST fire.
+  EXPECT_TRUE(activity_b.memoized_rank_fallback());
 
   // Both must produce the same VID — non-matching assignments behave exactly
   // like empty assignments.
--- a/src/test/cc/wfa/virtual_people/core/model/ranked_population_node_impl_test.cc
+++ b/src/test/cc/wfa/virtual_people/core/model/ranked_population_node_impl_test.cc
@@ -381,12 +381,11 @@ TEST(RankedPopulationNodeImplTest, MultipleRankAssignmentsResolvesCorrectPool) {
   EXPECT_LT(vid, 900);
 }
 
-// A fingerprint can legitimately route to multiple subpools across impressions
-// (design ÃÂ§ Cumulative rank index map). Phase-2 attaches every per-subpool rank
-// the fingerprint holds; this leaf takes its own pool_offset. If none matches
-// Ã¢ÂÂ because the fingerprint either overflowed this subpool in Phase 1 or
-// reached this subpool only on this impression and not in any previously
-// dispatched one Ã¢ÂÂ the leaf must fall back to the hash path, not fail the job.
+// When this leaf's pool_offset matches no entry in rank_assignments, the leaf
+// must fall back to the hash path instead of failing the job. The dominant
+// case is Phase-1 overflow: when this subpool reached ranked_size and this
+// fingerprint was one of the unranked surplus, the subpool has no rank for it
+// (per design § Retention Rule, overflow-fps-fall-back-to-unranked-path).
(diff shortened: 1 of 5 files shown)


Marco-Premier (approved):
> @Marco-Premier made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/89)* status: 0 of 5 files reviewed, 2 unresolved discussions (waiting on stevenwarejones).

___
*[`src/test/cc/wfa/virtual_people/core/model/ranked_population_node_impl_test.cc` line 393 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/89#-OvxWXN5CusvY9z64y5U:-OvxWXN5CusvY9z64y5V:b-aksa6z) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/6e803b89f464fe3a6cb3a33672e995ee65f4ed96/src/test/cc/wfa/virtual_people/core/model/ranked_population_node_impl_test.cc#L393)):*
> ```cpp
> // fingerprint was one of the unranked surplus, the subpool has no rank for it
> // (per design § Retention Rule, overflow-fps-fall-back-to-unranked-path).
> TEST(RankedPopulationNodeImplTest, NoMatchingRankAssignmentFallsBackToHash) {
> ```

NIT: Same comment for the kotlin test: `src/test/kotlin/org/wfanet/virtualpeople/core/labeler/RankedLabelingIntegrationTest.kt:251`

`NoMatchingRankAssignmentFallsBackToHash` verifies the hash VID range but never checks `memoized_rank_fallback`, so the flag is only proven on the boundary case — add `EXPECT_TRUE(activity.memoized_rank_fallback());` here (and `assertTrue(output.peopleList[0].memoizedRankFallback)` in the Kotlin test), since surfacing that signal on the no-match/overflow path is the whole point of the field.
___
*[`src/test/cc/wfa/virtual_people/core/model/ranked_population_node_impl_test.cc` line 438 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/89#-OvxXHng2cVCvcL08hUd:-OvxXHng2cVCvcL08hUe:b-9owcnt) ([raw file](https://github.com/world-federation-of-advertisers/virtual-people-core-serving/blob/6e803b89f464fe3a6cb3a33672e995ee65f4ed96/src/test/cc/wfa/virtual_people/core/model/ranked_population_node_impl_test.cc#L438)):*
> ```cpp
> // introduce a different hash seed for the two no-rank cases.
> TEST(RankedPopulationNodeImplTest,
>      NoMatchingRankAssignmentMatchesEmptyAssignmentsVid) {
> ```

NIT: Same comment for `RankedLabelingIntegrationTest.kt:297`:

The determinism test (`...MatchesEmptyAssignmentsVid`) is the ideal place to lock in both polarities — assert `EXPECT_FALSE(event_a…memoized_rank_fallback()`) (no assignments) and `EXPECT_TRUE(event_b…memoized_rank_fallback()`) (non-matching) so a future regression can't silently set/clear it.

src/test/cc/wfa/virtual_people/core/model/ranked_population_node_impl_test.cc
```diff
@@ -381,15 +385,20 @@ TEST(RankedPopulationNodeImplTest, MultipleRankAssignmentsResolvesCorrectPool) {
   EXPECT_LT(vid, 900);
 }
 
-TEST(RankedPopulationNodeImplTest, RankForNonExistentPoolReturnsError) {
+// When this leaf's pool_offset matches no entry in rank_assignments, the leaf
+// must fall back to the hash path instead of failing the job. The dominant
+// case is Phase-1 overflow: when this subpool reached ranked_size and this
+// fingerprint was one of the unranked surplus, the subpool has no rank for it
+// (per design § Retention Rule, overflow-fps-fall-back-to-unranked-path).
+TEST(RankedPopulationNodeImplTest, NoMatchingRankAssignmentFallsBackToHash) {
```
stevenwarejones, line 393: Done — `NoMatchingRankAssignmentFallsBackToHash` now asserts `EXPECT_TRUE(activity.memoized_rank_fallback())` on the C++ side (line 432) and `assertTrue(activity.memoizedRankFallback)` in the Kotlin `rank for non-existent pool falls back to hash` test (line 291).

src/test/cc/wfa/virtual_people/core/model/ranked_population_node_impl_test.cc
```diff
@@ -399,15 +408,81 @@ TEST(RankedPopulationNodeImplTest, RankForNonExistentPoolReturnsError) {
...
+  EXPECT_TRUE(activity.memoized_rank_fallback());
+}
+
+// The non-matching-rank_assignments path must produce the same VID as the
+// empty-rank_assignments path — both represent the same semantic state
+// (no rank for this fingerprint in this subpool, hash-fall-back), so same
+// input must yield the same VID. Locks in that the fix didn't accidentally
+// introduce a different hash seed for the two no-rank cases.
+TEST(RankedPopulationNodeImplTest,
+     NoMatchingRankAssignmentMatchesEmptyAssignmentsVid) {
```
stevenwarejones, line 442: Done — `NoMatchingRankAssignmentMatchesEmptyAssignmentsVid` now locks in both polarities: `EXPECT_FALSE(activity_a.memoized_rank_fallback())` for the empty-assignments path (line 468) and `EXPECT_TRUE(activity_b.memoized_rank_fallback())` for the non-matching path (line 481). Same polarity assertions added to the Kotlin `non-matching rank assignment produces same VID as empty rank assignments` test (lines 327, 341).

stevenwarejones (commented):
> @stevenwarejones resolved 4 discussions.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/89)* status: 0 of 5 files reviewed, all discussions resolved (waiting on raimundoltdf).

stevenwarejones (commented):
> @stevenwarejones partially reviewed 1 file.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/89)* status: 0 of 5 files reviewed, all discussions resolved (waiting on raimundoltdf).

raimundoltdf (approved):

stevenwarejones (commented):
> @stevenwarejones reviewed 6 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/89)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on stevenwarejones).

PR #91, opened 2026-07-17
build: git_override virtual-people-common for LabelerInput.placement

stevenwarejones:
Adds a git_override pinning virtual-people-common to the branch that introduces the LabelerInput.placement field (world-federation-of-advertisers/virtual-people-common#76), so this repo builds against the new field before it is released to the registry. The lockfile is regenerated to match. This is temporary plumbing: once the common change is released, remove the override and bump the registry version.

No functional change is needed in this repo for a model to route on the new field. The field filter engine lives in common (field_filter/utils/field_util.cc), and GetFieldFromProto resolves a dotted path such as labeler_input.placement generically through protobuf reflection (FindFieldByName plus GetReflection), reading the leaf value the same way. Any field reachable by dot path from LabelerEvent becomes usable in an EQUAL, IN, or REGEXP filter as soon as it exists in the compiled descriptor, with no per-field code. This override therefore only exposes the new descriptor to this repo's build; it does not wire up any behavior.

A natural follow-up is a labeler integration test that branches on labeler_input.placement into different VID pools and asserts that an Instagram placement and a Facebook placement land in different pools.


dawn-wang22 (approved):

dawn-wang22 (approved):

laureanobrs (approved):
> @laureanobrs reviewed 12 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/virtual-people-core-serving/91)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on stevenwarejones).

